PR #73, opened 2025-02-05
Adds missing pdiplugin-mpi dependency on Adastra

etiennemlb:
Adds missing pdiplugin-mpi dependency on Adastra.

This builds fine on Adastra.


EmilyBourne:
@etiennemlb I have temporarily made the GPU tests optional so we can merge this fork. Please use branches instead of forks for future developments so you are able to use all of the CI

EmilyBourne:
> > @etiennemlb I will temporarily make the GPU tests optional so we can merge this fork. Please use branches instead of forks for future developments so you are able to use all of the CI
> 
> Ok noted, did I miss other guidelines regarding the gyselalibxx GitHub workflow ? I simply tried to follow the "typical GitHub way of doing things".

No I don't think you missed anything else. The "typical GitHub way" should be the right way, but we have to leave the GPU CI for members of gyselax only to protect persee. This precludes the use of forks by members of gysela.

PR #84, opened 2025-02-07
Readme src/mapping

PaulineVidal:
Add more formulae in the README.

### DIFF
--- a/src/mapping/README.md
+++ b/src/mapping/README.md
@@ -48,7 +48,7 @@ with $\det(J) = r$.
 \right.
 ```
 
-with $ \xi = 1/\sqrt{1 - \varepsilon^2 /4}$.
+with $\xi = 1/\sqrt{1 - \varepsilon^2 /4}$.
 
 ```math
 J (r,\theta) 
--- a/src/mapping/README.md
+++ b/src/mapping/README.md
@@ -4,20 +4,120 @@ This folder contains code describing tools for handling different coordinate sys
 
 The current coordinate transformations implemented are:
 
-- Analytically invertible coordinate transformations such as:
-	-  Circular coordinate transformation (CircularToCartesian/CartesianToCircular):
-		-  $x(r,\theta) = r \cos(\theta),$
-	 	-  $y(r,\theta) = r \sin(\theta).$
-	-  Czarny coordinate transformation (CzarnyToCartesian/CartesianToCzarny):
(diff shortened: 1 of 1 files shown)


src/mapping/README.md
```diff
@@ -4,20 +4,120 @@ This folder contains code describing tools for handling different coordinate sys
...
+J (r,\theta) = 
+\begin{bmatrix} 
+	\cos(\theta) & -r\sin(\theta) \\
+	\sin(\theta) &  r\cos(\theta) \\
+\end{bmatrix}.
+```
+
+with $\det(J) = r$.
+
+- Inverse mapping (CartesianToCircular):
```
EmilyBourne, line 30: Do you think the terms "Mapping" and "Inverse mapping" are still pertinent? You could reorder these 2 sections and everything would still be accurate right?

PaulineVidal: If the mappings are functions, they have a starting domain and an arrival domain. 
image

If we reorder, we will talk about F^-1 and not F.

EmilyBourne: I agree but nothing in this README refers to the article that you have screenshotted. I don't think it would be invalid to say:

> We denote by $\Omega=[0,1]x[0,2\pi[$ the logical domain and by $R^2$ the physical domain. `CartesianToCircular` represents a mapping $F:R^2\rightarrow\Omega$

I want to remain as generic as possible in sections of the code that are not specific to a given problem. I think that is the case here. These mappings can be used for the advection on the poloidal plane $(r_{minor},\theta)$, but they would also be usable for the toroidal plane $(r_{major}, \phi)$ and possibly for other things. Are we certain that $F:\Omega\rightarrow R^2$ will be the default in all these cases? And if not is there a more generic way to talk about this without arbitrarily labelling one or the other as an "inverse mapping"?

PaulineVidal: Well, I think that when we talk about "circular mapping", most of people thinks of the mapping as described here, so $\Omega \rightarrow \mathbb{R}^2$. It could be surprising to say that the circular mapping refers to the following function $(x, y) \mapsto (r = \sqrt{x^2 + y^2}, \theta = atan(y, x))$.

alex-m-h: I think it looks good. The circular or polar coordinate transformation or mapping refers the usual mapping
$(r,\theta) \mapsto (r \cos \theta, r \sin \theta)$. This is the default that everyone knows so I would mention it first for sure and then describe `CartesianToCircular` as the inverse of that.

EmilyBourne: ok. Fine for me then :+1:

EmilyBourne (approved):

PR #81, opened 2025-02-07
Add tensor type

EmilyBourne:
Add a type to describe tensor objects. Fixes #75 .

This PR is (almost) equivalent to https://gitlab.maisondelasimulation.fr/gysela-developpers/gyselalibxx/-/merge_requests/852

PR #85, opened 2025-02-07
Use new vector type in existing code

EmilyBourne:
Use the new vector type that arises from #81 in the existing code.

This PR must be merged after #81


EmilyBourne:
It seems that this implementation may be slower than the use of Coordinate/TaggedVector. Maybe bad luck on persee, maybe related to the comments on previous PRs

src/data_types/tensor.hpp
```diff
@@ -151,6 +167,24 @@ class Tensor
...
+     * @brief A copy operator.
+     * @param coord The coordinate to be copied into the vector.
+     * @return A reference to the current tensor.
+     */
+    template <class... Dims>
+    KOKKOS_FUNCTION Tensor& operator=(Coord<Dims...> coord)
+    {
+        static_assert(
+                rank() == 1,
+                "Filling the tensor on initialisation is only permitted for 1D vector objects");
```
yanmnc, line 180: Copying a coordinate into a tensor is only possible for 1D tensor objects

src/mapping/vector_mapper.hpp
```diff
@@ -84,7 +84,7 @@ class VectorMapper<NDTag<XIn, YIn>, NDTag<XOut, YOut>, Mapping, ExecSpace>
                         Matrix_2x2 map_J;
                         mapping_proxy.jacobian_matrix(ddc::coordinate(idx), map_J);
 
-                        vector_element_type_out vector_out;
```
yanmnc, line 87: is this type used elsewhere in the code ? If no, its definition line 41 can be removed (same remark for vector_element_type_in

src/data_types/tensor.hpp
EmilyBourne: Done [902f8b1](https://github.com/gyselax/gyselalibxx/pull/85/commits/902f8b1fa083ba726d98ee0799c72bcaf9472a52)

src/mapping/vector_mapper.hpp
EmilyBourne: I agree that this is useless. Removed 2dedf9f

yanmnc (approved):

PR #88, opened 2025-02-11
Port the geometryRTheta interpolator to gpu

EmilyBourne:
Port the RTheta interpolator to GPU. For now it remains specific to the r,theta geometry.

### DIFF
--- a/src/geometryRTheta/interpolation/spline_interpolator_2d_rp.hpp
+++ b/src/geometryRTheta/interpolation/spline_interpolator_2d_rp.hpp
@@ -76,12 +76,15 @@ public:
      */
     host_t<DFieldRTheta> operator()(
             host_t<DFieldRTheta> const inout_data_host,
-            host_t<Field<CoordRTheta const, IdxRangeRTheta>> const coordinates_host) const override
+            host_t<ConstField<CoordRTheta, IdxRangeRTheta>> const coordinates_host) const override
     {
-        auto inout_data = ddc::
+        auto inout_data_alloc = ddc::
                 create_mirror_view_and_copy(Kokkos::DefaultExecutionSpace(), inout_data_host);
-        auto coordinates = ddc::
+        auto coordinates_alloc = ddc::
                 create_mirror_view_and_copy(Kokkos::DefaultExecutionSpace(), coordinates_host);
+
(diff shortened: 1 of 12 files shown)


src/geometryRTheta/interpolation/spline_interpolator_2d_rp.hpp
```diff
@@ -71,30 +68,29 @@ class SplineInterpolatorRTheta : public IInterpolatorRTheta
...
-                          == inout_data(r_idx_range.front(), theta_idx_range.front());
-                if (!unicity_center_point) {
-                    std::printf("Unicity of the value at the center point is not verified.");
-                    assert(unicity_center_point);
-                }
-            });
+            ddc::parallel_for_each(
+                    theta_idx_range,
+                    KOKKOS_LAMBDA(IdxTheta const ip) {
+                        KOKKOS_ASSERT(
```
gdgirard, line 84: There is no way to add a message error when using KOKKOS_ASSERT ?

src/math_tools/math_tools.hpp
```diff
@@ -64,7 +64,7 @@ constexpr inline double ipow(double a, std::size_t i)
     return r;
 }
 
-inline double ipow(double a, int i)
+KOKKOS_INLINE_FUNCTION double ipow(double a, int i)
```
gdgirard, line 67: That means that we are sure that ipow will always be called on GPU ?

EmilyBourne: No. `KOKKOS_INLINE_FUNCTION` is equivalent to `inline __device__ __host__` so this means that we can call `ipow` from **both** GPU and CPU

src/geometryRTheta/interpolation/spline_interpolator_2d_rp.hpp
EmilyBourne: No. `Kokkos::abort` allows us to specify an error message but `Kokkos::abort` also runs in debug mode. That said given that we are in a `#ifndef NDEBUG` region here maybe this doesn't matter. I will change this

EmilyBourne: Fixed 52f4b80

gdgirard (approved):

PR #91, opened 2025-02-12
Be more explicit about clang-format version

### LINKED ISSUE #90: Clang-format version in bin/indent.
In `bin/indent`, the called `clang-format` is the latest version. 
```
CLANG_FORMAT=clang-format
if [ "$(hostname)" == "persee.partenaires.cea.fr" ]
then
    CLANG_FORMAT=/data/gyselarunner/clang-format
fi
```

However, we use the version 14 of clang-format, `clang-format-14`.  So, for the users not coding on Persée, I suggest to change to 
```
CLANG_FORMAT=clang-format-14
if [ "$(hostname)" == "persee.partenaires.cea.fr" ]
then
    CLANG_FORMAT=/data/gyselarunner/clang-format
fi
```

Otherwise, the `indent` file will run the latest version of `clang-format` (version 18 or 19 maybe).

EmilyBourne:
Be more explicit about clang-format version to allow people to have multiple versions of clang-format installed on their computer. Fixes #90

### DIFF
--- a/bin/indent
+++ b/bin/indent
@@ -69,7 +69,7 @@ else
     RUN_IF_OK='true'
 fi
 
-CLANG_FORMAT=clang-format
+CLANG_FORMAT=clang-format-14
 if [ "$(hostname)" == "persee.partenaires.cea.fr" ]
 then
     CLANG_FORMAT=/data/gyselarunner/clang-format
--- a/bin/indent
+++ b/bin/indent
@@ -69,10 +69,18 @@ else
     RUN_IF_OK='true'
 fi
 
-CLANG_FORMAT=clang-format-14
 if [ "$(hostname)" == "persee.partenaires.cea.fr" ]
 then
     CLANG_FORMAT=/data/gyselarunner/clang-format
+else
+    CLANG_FORMAT=clang-format-14
+    set +e
+    if ! command -v ${CLANG_FORMAT} 2>&1 >/dev/null
+    then
+        CLANG_FORMAT=clang-format
+        echo "Could not find clang-format v14. Clang-formatting may not behave as expected"
+    fi
+    set -e
 fi
 
 ${DOCKER_COMMAND} find \


gdgirard:
I am still interested in a way to perform the indentation on my laptop without having to connect on Persee just for that. So if any solution, I'm interested in and it could be good to add a documentation on it somewhere.

EmilyBourne:
> I am still interested in a way to perform the indentation on my laptop without having to connect on Persee just for that. So if any solution, I'm interested in and it could be good to add a documentation on it somewhere.

@PaulineVidal managed to install the expected version of clang-format using:
```
sudo apt install clang-format-14
```

If you can confirm that this solution (possibly combined with this branch) works for you then I will add it to the docs

gdgirard:
Sorry I haven't looked at issue description. This should solve my problem. I am going to test it.

EmilyBourne:
> I've just tried on my computer I have the following message `find: ‘clang-format-14’: Aucun fichier ou dossier de ce nom` When I look at which clang is called I have
> 
> ```
> clang+llvm-18.1.7-x86_64-linux-gnu-ubuntu-18.04/bin/clang
> ```

You need the correct version of clang to get the correct indentation. Can you try installing the correct version of clang-format either using Pauline's apt command or via spack. I think the command should be something like
```
spack install clang-format@14
```

bin/indent
```diff
@@ -69,10 +69,18 @@ else
     RUN_IF_OK='true'
 fi
 
-CLANG_FORMAT=clang-format
 if [ "$(hostname)" == "persee.partenaires.cea.fr" ]
 then
     CLANG_FORMAT=/data/gyselarunner/clang-format
```
tpadioleau, line 74: Is it ok to expose this ? @gdgirard

gdgirard: Not sure that it is a good idea.

tpadioleau (approved):
Looks ok to me, maybe we should remove the Persee specifics in a future PR ?

PR #95, opened 2025-02-15
Fix CMAKE_TOOLCHAIN_FILE cmake variable

tpadioleau:
The toolchain is likely not used, there is a warning:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_TOOLCHAIN
```

### DIFF
--- a/.github/actions/build_code/action.yml
+++ b/.github/actions/build_code/action.yml
@@ -10,7 +10,7 @@ runs:
   steps:
     - name: Compile code
       run: |
-        cmake -DCMAKE_TOOLCHAIN=${{ inputs.toolchain }} -B build -S .
+        cmake -DCMAKE_TOOLCHAIN_FILE=${{ inputs.toolchain }} -B build -S .
         cmake --build build --parallel 4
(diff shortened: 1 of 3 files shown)


tpadioleau:
We likely want to run the tests as the toolchains may trigger some errors.

tpadioleau:
I don't know why the error for the failing test suddenly increased 🤔

EmilyBourne:
I'm not sure but I am having problems with the same test on my branches

EmilyBourne:
The tolerance is extremely low (1e-15) maybe a change to a dependency could have caused this? The test uses the splines from DDC but not much else

tpadioleau:
We recently removed the Kokkos Tools, we also use Kokkos installed from spack. But for those changes all tests passed. This is curious.

That said it seems still reasonable to slightly increase the tolerance.

EmilyBourne (approved):

EmilyBourne:
> Are they in sync with main ?

The ones that are passing (eg this one after syncing) are in sync with main

PR #97, opened 2025-02-17
CI improvements

EmilyBourne:
Minor improvements to the CI:
- Add a clean up workflow to take care of adding the `Ready to review` flag. This (should) ensure that this step is correctly triggered and takes into consideration both the CPU and GPU tests
- Remove unnecessary permissions (default are sufficient)
- Add missing `GH_TOKEN` to be able to correct flags
- Match SHA mentioned in commit to SHA used for GPU tests to make it easier to locate tests on GitLab
- Remove label workarounds in tests.yml that were designed for private forks but don't work (because private forks have to pay for CI)
- Remove concurrency limits

### DIFF
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -55,7 +55,7 @@ jobs:
           git config --global user.email "username@users.noreply.github.com"
           echo "${GITHUB_REPOSITORY}" > repo
           echo "${{ inputs.SHA || github.sha }}" > commit
-          git commit -m "Test commit ${GITHUB_SHA}" repo commit
+          git commit -m "Test commit ${{ inputs.SHA || github.sha }}" repo commit
           git push
         shell: bash
         env:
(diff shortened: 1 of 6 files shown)


.github/workflows/gpu_tests.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           git config --global user.email "username@users.noreply.github.com"
           echo "${GITHUB_REPOSITORY}" > repo
           echo "${{ inputs.SHA || github.sha }}" > commit
-          git commit -m "Test commit ${GITHUB_SHA}" repo commit
+          git commit -m "Test commit ${{ inputs.SHA || github.sha }}" repo commit
```
tpadioleau, line 58: Are `GITHUB_SHA` and `github.sha` the same ?

tpadioleau (approved):
lgtm

.github/workflows/gpu_tests.yml
EmilyBourne: Yes they should be. `github.sha` is calculated by github when it runs the commands `GITHUB_SHA` is an environment variable

PR #104, opened 2025-02-18
New aliases for SplineBuilder and SplineEvaluator definition

gdgirard:


src/utils/ddc_aliases.hpp
```diff
@@ -99,3 +100,328 @@ class Vector : public ddc::detail::TaggedVector<ElementType, Dims...>
...
+ */
+template <
+        class ExecSpace,
+        class BSplines,
+        class InterpolationGrid,
+        ddc::BoundCond BCLBound,
+        ddc::BoundCond BCUBound,
+        ddc::SplineSolver SolverType,
+        class IdxRangeType>
+class GetSplineBatchedBuilder1D;
```
EmilyBourne, line 120: I think these classes should be confined to a namespace (e.g. `namespace details{`) as they only exist as a tool to allow us to define the type aliases (e.g. `get_spline_batched_builder1d_t`).

EmilyBourne (approved):
Looks good to me

PR #108, opened 2025-02-21
Add a debug build type

tpadioleau:


tpadioleau:
@EmilyBourne Can we trigger the CI for modifications of toolchains ?

EmilyBourne:
@tpadioleau Yes feel free to change the workflow to trigger this correctly

.github/workflows/tests.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         with:
           # All of these options are optional, so you can remove them if you are happy with the defaults
           skip_after_successful_duplicate: 'true'
-          paths: '["src/**/*.hpp", "src/**/*.cpp", "tests/**/*.hpp", "tests/**/*.cpp", "simulations/**/*.hpp", "simulations/**/*.cpp", "**/CMakeLists.txt", "vendor/**", ".github/**"]'
+          paths: '["src/**/*.hpp", "src/**/*.cpp", "tests/**/*.hpp", "tests/**/*.cpp", "simulations/**/*.hpp", "simulations/**/*.cpp", "**/CMakeLists.txt", "vendor/**", ".github/**", "toolchains/**"]'
```
EmilyBourne, line 28: Should this be limited to the toolchains relevant for the CI? E.g. so we don't need to rerun tests if we update the Adastra toolchain

toolchains/docker.gyselalibxx_env/tests_toolchain.cmake
```diff
@@ -1,6 +1,9 @@
 
 include(${CMAKE_CURRENT_LIST_DIR}/../cicd_default_toolchain.cmake)
 
+# CMake options
+set(CMAKE_BUILD_TYPE FakeDebug)
```
EmilyBourne, line 5: Why `FakeDebug` instead of real `Debug`?

.github/workflows/tests.yml
tpadioleau: I agree, will udpate

toolchains/docker.gyselalibxx_env/tests_toolchain.cmake
tpadioleau: I was not sure we wanted `-g` but I am fine with `Debug`

.github/workflows/tests.yml
tpadioleau: Should we also include Persee toolchains ?

toolchains/docker.gyselalibxx_env/tests_toolchain.cmake
tpadioleau: Done in 769769005651c0031ce9aff248b068fedac82ced

.github/workflows/tests.yml
tpadioleau: Done in 769769005651c0031ce9aff248b068fedac82ced

EmilyBourne: > Should we also include Persee toolchains ?

The GPU one yes I think

PR #109, opened 2025-02-21
Allow tensors of references

EmilyBourne:
Represent tensor elements using `std::tuple` instead of `std::array` to allow `ElementType` to be a reference. Use vectors with references as elements to improve the use of `VectorField`

### DIFF
--- a/src/geometryRTheta/time_solver/bsl_predcorr_second_order_implicit.hpp
+++ b/src/geometryRTheta/time_solver/bsl_predcorr_second_order_implicit.hpp
@@ -253,16 +253,8 @@ public:
 
             // Compute the new advection field (E^n(X^n) + E^n(X^P)) /2:
             ddc::for_each(grid, [&](IdxRTheta const irp) {
-                ddcHelper::get<X>(advection_field_k_tot)(irp)
-                        = (ddcHelper::get<X>(advection_field)(irp)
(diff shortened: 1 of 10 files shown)


src/geometryRTheta/time_solver/bsl_predcorr_second_order_implicit.hpp
```diff
@@ -405,12 +390,7 @@ class BslImplicitPredCorrRTheta : public ITimeSolverRTheta
 
             // Compute the new advection field A(X^n) + A(X^{k-1}):
             ddc::for_each(grid, [&](IdxRTheta const irp) {
-                ddcHelper::get<X>(advection_field_k_tot)(irp)
-                        = ddcHelper::get<X>(advection_field)(irp)
-                          + ddcHelper::get<X>(advection_field_k)(irp);
-                ddcHelper::get<Y>(advection_field_k_tot)(irp)
-                        = ddcHelper::get<Y>(advection_field)(irp)
-                          + ddcHelper::get<Y>(advection_field_k)(irp);
+                advection_field_k_tot(irp) = advection_field(irp) + advection_field_k(irp);
```
EmilyBourne, line 393: Before this PR this line compiled. However no values were modified. Rather the temporary vector created in `operator()` was modified but did not affect the values saved in `advection_field_k_tot`

EmilyBourne:
Tuples do not work on GPU

PR #112, opened 2025-02-24
Improve compilation error output

### LINKED ISSUE #111: Missing compilation error when manipulating VectorFields
The result of a call to `VectorField::operator()` is a `Vector` object. However this object is created from copies of the elements of the vectors. This happens because the memory returned from `VectorField` is non-contiguous. One solution would be to use `std::tuple` in `Tensor` and allow the `ElementType` to be a reference. This was tried in #109 however `std::tuple` methods are not supported on GPU. For now we have to live with this restriction. This prevents us from doing things like:
```
void update_element(host_t<VectorField<double, IdxRangeXY, NDTag<X, Y>>> my_vec_field, DVector<X, Y> new_val, IdxXY pos) {
    my_vec_field(pos) = new_val;
}
```
However currently this code does not raise an error. It simply modifies the temporary `Vector` that was created, leaving `my_vec_field` unchanged. Instead `VectorField::operator()` should return a const object to ensure that an error is raised to avoid accidental bugs when writing code like this.

EmilyBourne:
Fix missing compilation error when manipulating VectorFields. Fixes #111

### DIFF
--- a/src/data_types/vector_field.hpp
+++ b/src/data_types/vector_field.hpp
@@ -184,8 +184,9 @@ private:
      * @return copy of this element
      */
     template <class... ODDims, typename T, T... ints>
-    KOKKOS_FUNCTION element_type const
-    operator()(Idx<ODDims...> const& delems, std::integer_sequence<T, ints...>) const noexcept
+    KOKKOS_FUNCTION element_type const operator()(
+            Idx<ODDims...> const& delems,
+            std::integer_sequence<T, ints...>) const noexcept
     {
         return element_type((base_type::m_values[ints](delems))...);
     }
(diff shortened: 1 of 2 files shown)


tpadioleau (approved):
I feel this is an anti-pattern. Clang-tidy has a check for that https://clang.llvm.org/extra/clang-tidy/checks/readability/const-return-type.html. I am not blocking though.

Maybe we should reconsider the design of the vector field instead ?

EmilyBourne:
> I feel this is an anti-pattern. Clang-tidy has a check for that https://clang.llvm.org/extra/clang-tidy/checks/readability/const-return-type.html. I am not blocking though.
> 
> Maybe we should reconsider the design of the vector field instead ?

I agree that it is not very clean. This solution was not my first choice (see issue description). However we are a bit stuck due to the fact that we need a VectorField to contain Fields so we can call the spline builder/evaluator on the elements associated with a dimension.

If you have any better ideas I am happy to discuss. In the meantime I will merge to prevent any unintended bugs

Issue #39, opened 2025-01-07
Where appropriate replace the use of `std::array` with `Vector`


EmilyBourne:
Where appropriate replace the use of `std::array` with `Vector` as defined here:
https://github.com/gyselax/gyselalibxx/blob/37a2d440c5f809169d096d3be9a2cfbe9ff38209/src/utils/ddc_aliases.hpp#L80-L101

This issue should be tackled after #38. All functions operating on Vectors should then include static_assertions to ensure that they are correctly mixing covariant/contravariant objects

EmilyBourne:
This seems to be completed

PR #113, opened 2025-02-24
Use Vector type in Polar Poisson solver 

EmilyBourne:
Part 1 of https://github.com/gyselax/gyselalibxx/issues/39

### DIFF
--- a/src/math_tools/math_tools.hpp
+++ b/src/math_tools/math_tools.hpp
@@ -22,10 +22,10 @@ KOKKOS_INLINE_FUNCTION T sum(const T* array, int size)
 
 template <class ElementType, class LayoutPolicy, class AccessorPolicy, std::size_t Ext>
 KOKKOS_INLINE_FUNCTION ElementType sum(Kokkos::mdspan<
-                       ElementType,
-                       Kokkos::extents<std::size_t, Ext>,
-                       LayoutPolicy,
-                       AccessorPolicy> const& array)
+                                       ElementType,
+                                       Kokkos::extents<std::size_t, Ext>,
+                                       LayoutPolicy,
+                                       AccessorPolicy> const& array)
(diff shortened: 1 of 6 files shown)


src/math_tools/math_tools.hpp
```diff
@@ -89,16 +91,14 @@ inline std::size_t factorial(std::size_t f)
     return r;
 }
 
-template <class T, std::size_t D>
-KOKKOS_INLINE_FUNCTION T dot_product(std::array<T, D> const& a, std::array<T, D> const& b)
+template <class T, class... Dims>
+KOKKOS_INLINE_FUNCTION T
+dot_product(Vector<T, Dims...> const& a, Vector<T, typename Dims::Dual...> const& b)
```
tpadioleau, line 96: Should this one be moved to the directory containing tensor.hpp in the future ?

We could make a distinction between math functions that launch a kernel from the ones that execute in the calling thread ?

tpadioleau (approved):

src/math_tools/math_tools.hpp
EmilyBourne: > Should this one be moved to the directory containing tensor.hpp in the future ?

I don't think so. The directory containing tensor.hpp contains data types. This is a mathematical function that happens to use a particular data type.

> We could make a distinction between math functions that launch a kernel from the ones that execute in the calling thread ?

I agree that this could be useful. I think this file should only contain functions that execute in the calling thread

tpadioleau: I mean we chose to define math operators (`operator+` and others) in `tensor.hpp`, the distinction with `dot_product` is not clear to me.

EmilyBourne: Ah good point

PR #117, opened 2025-02-25
Replace Matrix2x2 types with DTensor type for metric tensor computations

yanmnc:


src/mapping/metric_tensor_evaluator.hpp
```diff
@@ -44,51 +45,62 @@ class MetricTensor
...
      * with @f$ J_{\mathcal{F}} @f$ the Jacobian matrix.
      *
      * @param[in] coord
      * 				The coordinate where we evaluate the metric tensor.
-     * @param[out] matrix
-     * 				The metric tensor matrix.
+     * @return metric_tensor
+     * 				A DTensor object containing the value of the metric tensor.
      */
-    KOKKOS_FUNCTION void operator()(Matrix_2x2& matrix, PositionCoordinate const& coord) const
```
EmilyBourne, line 61: Is the definition of `Matrix_2x2` above still needed?

tests/mapping/metric_tensor_evaluator.cpp
```diff
@@ -107,20 +107,37 @@ using Matrix_2x2 = std::array<std::array<double, 2>, 2>;
...
-                id_val += matrix[i][k] * inv[j][k];
-            }
-            EXPECT_NEAR(id_val, static_cast<double>(i == j), TOL);
-        }
-    }
+
+    using Dim0 = ddc::type_seq_element_t<0, Dims>;
+    using Dim1 = ddc::type_seq_element_t<1, Dims>;
+    using Dim0_cov = typename Dim0::Dual;
+    using Dim1_cov = typename Dim1::Dual;
```
EmilyBourne, line 120:
```suggestion
    static_assert(ddc::type_seq_size_v<Dims> == 2);
    using Dim0 = ddc::type_seq_element_t<0, Dims>;
    using Dim1 = ddc::type_seq_element_t<1, Dims>;
    using Dim0_cov = typename Dim0::Dual;
    using Dim1_cov = typename Dim1::Dual;
```
?

src/mapping/metric_tensor_evaluator.hpp
```diff
@@ -14,12 +14,13 @@
  * @tparam PositionCoordinate The coordinate type where the metric tensor can be evaluated.
  */
 template <class Mapping, class PositionCoordinate>
-class MetricTensor
+class MetricTensorEvaluator
```
gdgirard, line 17: Why do you prefer `MetricTensorEvaluator` than `MetricTensor` ?

tests/geometryRTheta/advection_field_rp/advection_field_gtest.cpp
```diff
@@ -241,22 +241,25 @@ TEST(AdvectionFieldRThetaComputation, TestAdvectionFieldFinder)
     // > Compare the advection field computed on RTheta to the advection field computed on XY
     host_t<DVectorFieldMemRTheta<X, Y>> difference_between_fields_xy_and_rp(grid);
 
-    MetricTensor<LogicalToPhysicalMapping, CoordRTheta> metric_tensor(to_physical_mapping);
+    MetricTensorEvaluator<LogicalToPhysicalMapping, CoordRTheta> metric_tensor(to_physical_mapping);
     ddc::for_each(grid_without_Opoint, [&](IdxRTheta const irp) {
         CoordRTheta const coord_rp(ddc::coordinate(irp));
 
         std::array<std::array<double, 2>, 2> J; // Jacobian matrix
```
gdgirard, line 248: Is it normal to keep this line unchanged ?

EmilyBourne: Yes. This PR only changes the metric tensor calculations. The Jacobian will be updated to a tensor in another PR

src/mapping/metric_tensor_evaluator.hpp
yanmnc: It isn't, I removed it

tests/geometryRTheta/advection_field_rp/advection_field_gtest.cpp
yanmnc: I'm marking this as resolved

src/mapping/metric_tensor_evaluator.hpp
yanmnc: The class DTensor stores the values contained in the tensor it represents, but the MetrictensorEvaluator does not: it is only used through operator() and inverse() functions to compute the coefficients of the metric tensor and to store them in the DTensor returned by these functions. So by renaming the MetrictensorEvaluator class it is more obvious that it does not really represent a tensor, but rather serves as a utility class to compute a tensor. What do you think ?

EmilyBourne (approved):

PR #118, opened 2025-02-26
Update Persee environment with Kokkos 4.5.1, Kokkos Kernels 4.5.1 and Kokkos-fft 0.2.1

tpadioleau:
- [x] Rely on packages installed by spack on Persee
- [x] Update in consequence toolchains
- [x] Before merging, rename the environment `gyselalibxx-env-omp-cuda-candidate` to `gyselalibxx-env-omp-cuda`

### DIFF
--- a/toolchains/v100.persee/debug_toolchain.cmake
+++ b/toolchains/v100.persee/debug_toolchain.cmake
@@ -6,6 +6,7 @@ set(CMAKE_BUILD_TYPE Debug)
 
 # Compiler options
 set(CMAKE_CXX_COMPILER nvcc_wrapper)
+set(CMAKE_CXX_EXTENSIONS OFF) # Avoid a Kokkos warning that will force if to OFF anyway when compiling with nvcc
 set(CMAKE_CXX_FLAGS "-g -Wall -Werror -Wno-sign-compare -Xcudafe --diag_suppress=unsigned_compare_with_zero -Xcudafe --diag_suppress=integer_sign_change -Wno-unused-but-set-variable")
 
 # Activate/deactivate parts of the code
--- a/toolchains/v100.persee/debug_toolchain.cmake
+++ b/toolchains/v100.persee/debug_toolchain.cmake
@@ -5,13 +5,8 @@ include(${CMAKE_CURRENT_LIST_DIR}/../cicd_default_toolchain.cmake)
 set(CMAKE_BUILD_TYPE Debug)
 
 # Compiler options
-set(CMAKE_CXX_COMPILER ${CMAKE_CURRENT_LIST_DIR}/../../vendor/kokkos/bin/nvcc_wrapper)
+set(CMAKE_CXX_COMPILER nvcc_wrapper)
 set(CMAKE_CXX_FLAGS "-g -Wall -Werror -Wno-sign-compare -Xcudafe --diag_suppress=unsigned_compare_with_zero -Xcudafe --diag_suppress=integer_sign_change -Wno-unused-but-set-variable")
-
-# Kokkos options
-set(Kokkos_ENABLE_SERIAL ON CACHE BOOL "Allow serial code to run" FORCE)
-set(Kokkos_ENABLE_CUDA ON CACHE BOOL "Activate GPU usage via cuda" FORCE)
-set(Kokkos_ARCH_VOLTA70 ON CACHE BOOL "Indicate that the GPU architecture is V100" FORCE)
 
 # Activate/deactivate parts of the code
 if (DEFINED ENV{DDC_BUILD_TESTING})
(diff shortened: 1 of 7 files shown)


tpadioleau:
Shouldn't we rename `cicd_default_toolchain.cmake` ? It looks like only CI related toolchains use it but all toolchains are based on it. What about `base_toolchain.cmake` ?

EmilyBourne:
> Shouldn't we rename `cicd_default_toolchain.cmake` ? It looks like only CI related toolchains use it but all toolchains are based on it. What about `base_toolchain.cmake` ?

That might be a good idea. I wanted to use a name that would discourage people from using it directly.

tpadioleau:
> That might be a good idea. I wanted to use a name that would discourage people from using it directly.

Do you have any idea ? What about `incomplete_toolchain.cmake` ?

tpadioleau:
If that is still ok for you, I will update the tolerance of the test `JacobianMatrixAndJacobianCoefficients.MatrixDiscCzarMap` to `1E-13` ? One possible explanation for this difference of result is that the spack installation of Kokkos specifies the host architecture whereas we were not until now.

EmilyBourne:
> If that is still ok for you, I will update the tolerance of the test `JacobianMatrixAndJacobianCoefficients.MatrixDiscCzarMap` to `1E-13` ? One possible explanation for this difference of result is that the spack installation of Kokkos specifies the host architecture whereas we were not until now.

Ok for me

EmilyBourne:
> > That might be a good idea. I wanted to use a name that would discourage people from using it directly.
> 
> Do you have any idea ? What about `incomplete_toolchain.cmake` ?

I'm not sure. I don't like `incomplete_toolchain.cmake` but I can't work out why so this is probably fine

EmilyBourne:
> > I'm not sure. I don't like `incomplete_toolchain.cmake` but I can't work out why so this is probably fine
> 
> And compared to `base_toolchain.cmake` ? A preference ? :D

`base_toolchain.cmake` feels cleaner, but I think `incomplete_toolchain.cmake` is less likely to be misused

gdgirard:
> > > I'm not sure. I don't like `incomplete_toolchain.cmake` but I can't work out why so this is probably fine
> > 
> > 
> > And compared to `base_toolchain.cmake` ? A preference ? :D
> 
> `base_toolchain.cmake` feels cleaner, but I think `incomplete_toolchain.cmake` is less likely to be misused

Not sure to understand what you mean by "incomplete" and why you want to keep it in this case ?

gdgirard:
> > Not sure to understand what you mean by "incomplete" and why you want to keep it in this case ?
> 
> It does not set the Kokkos options if needed nor `CMAKE_BUILD_TYPE`. It is likely to not give what people want.

And in which case will we need this environment ?

tpadioleau:
> And in which case will we need this environment ?

Never I think, this is a helper to factorize some cmake code common to different toolchains

tpadioleau:
@EmilyBourne @gdgirard May I get an approval for the GPU tests ?

EmilyBourne:
> I suggest we postpone this discussion, otherwise we will need to wait for the CI if we change the branch now ? I can open an issue about this if you want.

That was why I didn't trigger the GPU CI :wink: the CPU tests are slower than the GPU tests anyway so if we can come to a conclusion quickly then there will be no extra wait time. Otherwise I am happy to leave this for later.

Did anyone have an opinion on `importable_base_toolchain.cmake` ? `importable` to indicate that the toolchain should be imported from another

tpadioleau:
> Did anyone have an opinion on `importable_base_toolchain.cmake` ? `importable` to indicate that the toolchain should be imported from another

I feel it has the same issue with `base`, it is not clear it will not work ? What about prefixing it with a `.` to make it hidden ?

EmilyBourne:
I have created a dedicated discussion:
https://github.com/gyselax/gyselalibxx/discussions/123

tpadioleau:
Should be good now, the `spack env rename` did not regenerate the view.

tpadioleau:
urgh timeouts, Persee is too busy

toolchains/v100.persee/debug_toolchain.cmake
```diff
@@ -5,14 +5,10 @@ include(${CMAKE_CURRENT_LIST_DIR}/../cicd_default_toolchain.cmake)
 set(CMAKE_BUILD_TYPE Debug)
 
 # Compiler options
-set(CMAKE_CXX_COMPILER ${CMAKE_CURRENT_LIST_DIR}/../../vendor/kokkos/bin/nvcc_wrapper)
+set(CMAKE_CXX_COMPILER nvcc_wrapper)
+set(CMAKE_CXX_EXTENSIONS OFF) # Avoid a Kokkos warning that will force if to OFF anyway when compiling with nvcc
```
EmilyBourne, line 9: Should `set(CMAKE_CXX_EXTENSIONS OFF)` be moved to `toolchains/cicd_default_toolchain.cmake` ? Or is it ON for CPU configs?

tpadioleau: Kokkos does not force it to `OFF` for any other compiler than `nvcc`. So I would keep it only for Nvidia GPU clusters using `nvcc`.

EmilyBourne (approved):

PR #120, opened 2025-02-26
Add a spell checker

yanmnc:


.github/workflows/tests.yml
```diff
@@ -98,3 +98,14 @@ jobs:
...
+  spelling:
+   name: Spell Check with Typos
+   runs-on: ubuntu-latest
+   steps:
+   - name: Checkout Actions Repository
+     uses: actions/checkout@v4
+   - name: Spell Check Repo
+     uses: crate-ci/typos@v1.29.10
+     with:
+       files: ./AUTHORS ./CMakeLists.txt ./docs/ ./README.md ./tests/ ./toolchains
```
EmilyBourne, line 111: I think this would be better in this file: https://github.com/gyselax/gyselalibxx/blob/main/.github/workflows/static_analyses.yml

with the static analysis tests that are not costly and are therefore always run (even when the PR is in draft)

EmilyBourne, line 111:
```suggestion
       files: ./AUTHORS ./CMakeLists.txt ./docs/ ./README.md ./tests/ ./toolchains ./src ./simulations
```
?

PR #127, opened 2025-02-26
Replace Matrix2x2 types with DTensor type for Jacobian computations 

EmilyBourne:
Use DTensor for the Jacobian and its inverse.


EmilyBourne:
As #114 is ready to review, it does not make sense to complete this PR by explicitly writing 2D inverse/determinant/mat_mul functions.

EmilyBourne:
Blocked by #175

src/geometryRTheta/advection/bsl_advection_rtheta.hpp
```diff
@@ -186,19 +187,15 @@ class BslAdvectionRTheta : public IAdvectionRTheta
...
-                                * jacobian;
-            ddcHelper::get<Y>(advection_field_xy_host)(irtheta)
-                    = ddcHelper::get<R_cov>(advection_field_rtheta)(irtheta) * inv_J[0][1]
-                              * jacobian
-                      + ddcHelper::get<Theta_cov>(advection_field_rtheta)(irtheta) * inv_J[1][1]
-                                * jacobian;
+            DVector<R_cov, Theta_cov> advec_field_rtheta = advection_field_rtheta(irtheta);
+            // advec_field_xy = J^{-T} A
+            DVector<X, Y> advec_field_xy
+                    = tensor_mul(index<'i', 'j'>(inv_J), index<'i'>(advec_field_rtheta)) * jacobian;
```
EmilyBourne, line 196: @PaulineVidal Why do we have:
$A_{xy} = J^{-T} A_{r\theta} |J|$
?

The |J| does not seem to be expected. Can I remove it?
image

PaulineVidal: It is part of the meeting we had yesterday. I made a mistake on the definition of $A_{r,\theta}$. 
We are supposed to have $A_{x,y} = J A_{r,\theta}$.

PaulineVidal: Sorry I realised my comment was apparently not sent. In the meantimes, I created the PR related to this issue: https://github.com/gyselax/gyselalibxx/pull/178

EmilyBourne: I don't think we have $A_{x,y} = J A_{r,\theta}$ here. $A$ is expressed on the covariant basis so we have $A^{x,y} = J^{-T} A^{r,\theta}$ as indicated in Edoardo's article (see screenshot above).

PaulineVidal: I am not sure, 
$$\partial_t f + A_{x,y}\cdot \nabla_{x,y} f  = 0$$
and in Edoardo's article, $\nabla_{x,y} f  = J^{-T} \nabla_{r,\theta} f $, so 
$$\partial_t f + A_{x,y}\cdot (J^{-T} \nabla_{r,\theta} f)  = 0$$
i.e.
$$\partial_t f + (J^{-1} A_{x,y})\cdot \nabla_{r,\theta} f  = 0$$
so, 
$$A_{r,\theta} = J^{-1} A_{x,y}$$ 
and 
$$A_{x,y} = J A_{r,\theta}$$

PaulineVidal: What are the difference between $\nabla_{r,\theta} \phi$ and $\nabla^{r,\theta} \phi$? What are the definition of both?

EmilyBourne: > we know that $G=J J^T$

except that it's $G=J^T J$ :thinking: I still think the equation in Edoardo's article is probably correct though

PaulineVidal: I dont say it is incorrect, I was questionning the definition of the advection field on (r,theta).

EmilyBourne: > What are the difference between $\nabla_{r,\theta} \phi$ and $\nabla^{r,\theta} \phi$?

I would say that in the first case we express the contravariant components, while in the second case we express the covariant components.

> What are the definition of both? 

I think (but am not super sure right now) that:
$\nabla^{r,theta} = (\partial_r, \partial_\theta)$
while:
$\nabla_{r,\theta} = G (\partial_r, \partial_\theta)$

yanmnc (changes requested):

yanmnc (approved):

PR #133, opened 2025-02-28
Add an omp toolchain for Persee

### LINKED ISSUE #131: Add a CPU toolchain on Persee
For the moment only GPU toolchains exist on persee. We need to use docker ones to compile on CPU, which seems to be inconsistent with the sourcing of the environment.

tpadioleau:
closes #131

cc @protaisM

### DIFF
--- a/toolchains/xeon.persee/debug_toolchain.cmake
+++ b/toolchains/xeon.persee/debug_toolchain.cmake
@@ -1,0 +1,14 @@
+
+include(${CMAKE_CURRENT_LIST_DIR}/../cicd_default_toolchain.cmake)
+
+# CMake options
+set(CMAKE_BUILD_TYPE Debug)
+
+# Compiler options
+set(CMAKE_CXX_FLAGS "-Wall -Wno-sign-compare -Wno-unused-but-set-variable")
+
+# Activate/deactivate parts of the code
+if (DEFINED ENV{DDC_BUILD_TESTING})
+    set(DDC_BUILD_TESTS $ENV{DDC_BUILD_TESTING})
+endif()
+set(ACTIVATE_RESTART_TESTS OFF)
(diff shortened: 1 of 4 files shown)


toolchains/xeon.persee/debug_toolchain.cmake
```diff
@@ -0,0 +1,14 @@
...
+set(CMAKE_BUILD_TYPE Debug)
+
+# Compiler options
+set(CMAKE_CXX_FLAGS "-Wall -Wno-sign-compare -Wno-unused-but-set-variable")
+
+# Activate/deactivate parts of the code
+if (DEFINED ENV{DDC_BUILD_TESTING})
+    set(DDC_BUILD_TESTS $ENV{DDC_BUILD_TESTING})
+endif()
+set(ACTIVATE_RESTART_TESTS OFF)
```
tpadioleau, line 14: I copy-pasted this part, I don't know if it should be added or not

EmilyBourne: Where did you copy/paste it from?

tpadioleau: From `v100.persee/toolchain.cmake` that actually should be called `release_toochain.cmake`.

EmilyBourne: Ok. I think the aim of this block was to use the environment variable set here:
https://github.com/gyselax/gyselalibxx/blob/main/.github/actions/test_filter/action.yml#L38-L39

to run GPU tests for DDC. I don't think it is relevant here

tpadioleau: I will remove it then

EmilyBourne: Ok. I have removed the "Ready to review" label for now. Please put it back when you have updated this

tpadioleau: done in 075d70a41aaf5550daa7cb021110d4a0d1b1e170 with renaming of `release_toolchain` to follow the convention of other toolchains.

EmilyBourne (approved):

PR #135, opened 2025-02-28
Make sure postproc/ folder complies with spell checker

yanmnc:
also adds this directory to the list of directories checked by the spell checker

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -6,6 +6,7 @@ extend-ignore-re = ["capitalize()",
                     "ddc::\\S*",
                     "gko_exec->\\S*",
                     "on_finalize:",
+                    "colors.Normalize"
                     ]
 
 [default.extend-words]
@@ -17,3 +18,4 @@ Ax = "Ax"
 ax = "ax"
 color = "color"
 colors = "colors"
+iit = "iit"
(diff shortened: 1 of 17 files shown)


.typos.toml
```diff
@@ -17,3 +18,4 @@ Ax = "Ax"
 ax = "ax"
 color = "color"
 colors = "colors"
+iit = "iit"
```
EmilyBourne, line 21: ?

yanmnc: There are a few occurence of the 'iit' variable used to index arrays, see for instance [here](https://github.com/gyselax/gyselalibxx/blob/2d546a9a0f3c1cddc7739b5c9f278b40aa311627/post-process/PythonScripts/diag_efield.py#L168). The spellchecker does not like this so either we keep this entry in the toml file, or we change the name of the iit index

EmilyBourne: I kind of agree with the spell checker. I don't find `iit` very understandable. I think it is the index of the time index? Is that right?

yanmnc: Yes that's how I understand it as well. I changed the name in f96280b6d662a908a4b31c7016406bc121470527

EmilyBourne (approved):

PR #144, opened 2025-02-28
Correct definition of r_extent and theta_extent in collision operator

gdgirard:


EmilyBourne (approved):
This looks ok to me as long as @etiennemlb agrees

Issue #67, opened 2025-01-31
Choose a better name for R the radial coordinate


EmilyBourne:
Choose a better name for R the radial coordinate. In equations it is usually referred to as `r` and `R` represents something else. However in C++ capitalisation is usually used for classes while lowercase letters are used for variable names

EmilyBourne:
After discussion with @gdgirard and @PaulineVidal we suggest `Rminor` and `Rmajor` in files where there may be confusion (e.g. simulations). In files which are specific to a known geometry `R` is permitted. E.g. `R` is permitted in a class handling a polar plane only

PR #145, opened 2025-03-01
Change the "p" into "theta"

PaulineVidal:
In the geometryRTheta a lot of "p" for polar remain in the files. I changed them into "theta" as we decided to use "theta" instead of "p".


PaulineVidal:
@EmilyBourne I created an issue for the `.uid()` functions in polarpoissonlikesolver.hpp : https://github.com/gyselax/gyselalibxx/issues/146

I don't understand the error 
```
style: Local variable 'idx_theta' shadows outer variable [shadowVariable]
                                IdxQuadratureTheta const idx_theta(idx_quad);
``` 
in the static analysis tests.

src/interpolation/polar_splines/polar_bsplines.hpp
```diff
@@ -321,10 +321,11 @@ class PolarBSplines
                         = ddc::discrete_space<BSplinesTheta>().full_domain();
 
                 for (IdxR const ir : IdxRange<BSplinesR>(IdxR(0), IdxStepR(C + 1))) {
-                    for (IdxTheta const ip : poloidal_spline_idx_range.take_first(np_in_singular)) {
+                    for (IdxTheta const itheta :
+                         poloidal_spline_idx_range.take_first(np_in_singular)) {
```
EmilyBourne, line 325:
```suggestion
                         poloidal_spline_idx_range.take_first(n_theta_in_singular)) {
```
?

tests/geometryRTheta/polar_poisson/polarpoissonfemsolver.cpp
```diff
@@ -81,24 +81,25 @@ int main(int argc, char** argv)
...
+    ddc::init_discrete_space<BSplinesTheta>(theta_knots);
 
     ddc::init_discrete_space<GridR>(SplineInterpPointsR::get_sampling<GridR>());
     ddc::init_discrete_space<GridTheta>(SplineInterpPointsTheta::get_sampling<GridTheta>());
 
     IdxRangeR interpolation_idx_range_R(SplineInterpPointsR::get_domain<GridR>());
-    IdxRangeTheta interpolation_idx_range_P(SplineInterpPointsTheta::get_domain<GridTheta>());
-    IdxRangeRTheta grid(interpolation_idx_range_R, interpolation_idx_range_P);
+    IdxRangeTheta interpolation_idx_range_Theta(SplineInterpPointsTheta::get_domain<GridTheta>());
+    IdxRangeRTheta grid(interpolation_idx_range_R, interpolation_idx_range_Theta);
```
EmilyBourne, line 102: Unusual to use capitals in variable names. According to our [coding standards](https://gyselax.github.io/gyselalibxx/docs_CODING_STANDARD.html#docs_CODING_STANDARD__Naming) we use `snake_case` for variables.
```suggestion
    IdxRangeTheta interpolation_idx_range_Theta(SplineInterpPointsTheta::get_domain<GridTheta>());
    IdxRangeRTheta grid(interpolation_idx_range_R, interpolation_idx_range_Theta);
```

tests/geometryRTheta/quadrature/tests_L1_and_L2_norms.cpp
```diff
@@ -201,23 +201,24 @@ TEST_P(SplineQuadrature, TestFunctions)
...
+    ddc::init_discrete_space<BSplinesTheta>(theta_knots);
 
     ddc::init_discrete_space<GridR>(SplineInterpPointsR::get_sampling<GridR>());
     ddc::init_discrete_space<GridTheta>(SplineInterpPointsTheta::get_sampling<GridTheta>());
 
     IdxRangeR interpolation_idx_range_R(SplineInterpPointsR::get_domain<GridR>());
-    IdxRangeTheta interpolation_idx_range_P(SplineInterpPointsTheta::get_domain<GridTheta>());
-    IdxRangeRTheta grid(interpolation_idx_range_R, interpolation_idx_range_P);
+    IdxRangeTheta interpolation_idx_range_Theta(SplineInterpPointsTheta::get_domain<GridTheta>());
+    IdxRangeRTheta grid(interpolation_idx_range_R, interpolation_idx_range_Theta);
```
EmilyBourne, line 221: Same comment as above
```suggestion
    IdxRangeTheta interpolation_idx_range_theta(SplineInterpPointsTheta::get_domain<GridTheta>());
    IdxRangeRTheta grid(interpolation_idx_range_r, interpolation_idx_range_theta);
```

tests/interpolation/polar_splines.cpp
```diff
@@ -139,9 +139,9 @@ TEST(PolarSplineTest, ConstantEval)
     ddc::init_discrete_space<GridR>(GrevillePointsR::get_sampling<GridR>());
     ddc::init_discrete_space<GridTheta>(GrevillePointsTheta::get_sampling<GridTheta>());
     IdxRange<GridR> interpolation_idx_range_R(GrevillePointsR::get_domain<GridR>());
-    IdxRange<GridTheta> interpolation_idx_range_P(GrevillePointsTheta::get_domain<GridTheta>());
+    IdxRange<GridTheta> interpolation_idx_range_Theta(GrevillePointsTheta::get_domain<GridTheta>());
     IdxRange<GridR, GridTheta>
-            interpolation_idx_range(interpolation_idx_range_R, interpolation_idx_range_P);
+            interpolation_idx_range(interpolation_idx_range_R, interpolation_idx_range_Theta);
```
EmilyBourne, line 144: Same comment
```suggestion
            interpolation_idx_range(interpolation_idx_range_r, interpolation_idx_range_theta);
```

tests/interpolation/polar_splines.cpp
```diff
@@ -262,9 +262,9 @@ void test_polar_integrals()
...
```
EmilyBourne, line 267: Same comment
```suggestion
            interpolation_idx_range(interpolation_idx_range_r, interpolation_idx_range_theta);
```

EmilyBourne (changes requested):

EmilyBourne:
Thank you for all these changes!

simulations/geometryRTheta/diocotron/CMakeLists.txt
```diff
@@ -29,10 +29,10 @@ function (diocotron_executable PREDCORR_METHOD TIME_METHOD )
 
             gslx::poisson_RTheta
             gslx::geometry_RTheta
-            gslx::interpolation_2D_rp
+            gslx::interpolation_polar
             gslx::io
-            gslx::advection_rp
-            gslx::time_integration_rp
+            gslx::advection_polar
+            gslx::time_integration_rtheta
```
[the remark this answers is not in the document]
PaulineVidal: Sorry I think I changed my mind in the middle. I forget theses files. 
But actually, what do you prefer between calling files "rtheta"  or "polar".

tests/geometryRTheta/polar_poisson/polarpoissonfemsolver.cpp
EmilyBourne: Yes I think so

src/interpolation/polar_splines/polar_bsplines.hpp
EmilyBourne: Unresolving this conversation as you seem to have applied the change without changing anything else but `np_in_singular` appears multiple times in this file and I don't think the declaration or other occurrences have been changed

simulations/geometryRTheta/diocotron/CMakeLists.txt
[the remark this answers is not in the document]
EmilyBourne: I don't have a strong preference but I did think that maybe `polar` was good for things that will be generalised out of the RTheta geometry. E.g. `TokamAxi` uses `Tor1` and `Tor2` for (R, Theta) so maybe polar is more intuitive for these cases (especially given the confusion around R as mentioned in https://github.com/gyselax/gyselalibxx/issues/67).

The question then is if you prefer to have polar for things that work on any polar geometry and rtheta for things that are specific to the rtheta geometry or if we just uniformise and put polar everywhere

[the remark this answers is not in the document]
PaulineVidal: The problem is that I dont really understand the difference between polar geometry and rtheta geometry. What is the difference? 

As the folder is called `geometryRThtea` I thought it should be better if we call the files and lib "rtheta" but I also realised that the Poisson solver is called "polar" and in the general advection folder, we called the advection "polar".

tests/geometryRTheta/polar_poisson/polarpoissonfemsolver.cpp
PaulineVidal: Done in https://github.com/gyselax/gyselalibxx/pull/145/commits/5ff6ecf11d07b6b7b74ea4520a94e923e4273a6d

simulations/geometryRTheta/diocotron/CMakeLists.txt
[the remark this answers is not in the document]
PaulineVidal: So, maybe as I'm working only in the folders `geomtryRTheta`, I should only use `rtheta`. And if we use work on files outside of this folder, we should use `polar` e.g. in `src/advection/` and `src/pde_solvers/`.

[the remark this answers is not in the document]
EmilyBourne: For me rtheta geometry is the geometry defined here:
https://github.com/gyselax/gyselalibxx/blob/main/src/geometryRTheta/geometry/geometry.hpp

It has a certain number of explicit and implicit assumptions that you use for the diocotron/vorticity/etc simulations.

A polar geometry is a set of coordinates $(\xi_1, \xi_2)$ (which may or may not be called "r" and "theta") which are defined such that $\xi_1$ describes the distance from an O-point and $\xi_2$ describes an angle.

For example in a tokamak I think that when **you** talk about geometryRTheta you have in mind the $(r,\theta)$ dimensions in this figure:
image


However I would say that there are 2 sets of coordinates that could be treated as a polar geometry. $(r,\theta)$ and $(R, \phi)$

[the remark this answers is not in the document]
PaulineVidal: Yes the geometryRTheta refers to the $(r,\theta)$ in the poloidal cross-section. 

Ok, so when you defined the `spline_polar_foot_finder.hpp` in `src/advection`  of the `polarpoissonlikesolver.hpp` in `src/pde_solvers`, can we also apply these operators to the plane with $(R,\varphi)$ coordinates?

src/interpolation/polar_splines/polar_bsplines.hpp
PaulineVidal: Done in https://github.com/gyselax/gyselalibxx/pull/145/commits/b23ad32d9a23980a37fcb8e71e261b5dea5f565c

simulations/geometryRTheta/diocotron/CMakeLists.txt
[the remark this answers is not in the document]
EmilyBourne: I would say yes in theory, although there is probably no use for this.

Maybe a system where this operator could theoretically make sense would be $(\rho_0, \zeta)$ ?
image

But my point was mainly that this operator can be applied to (tor1, tor2) not just (R, Theta) from the RTheta geometry. It could also be applied to a differently shaped machine which might be defined with a slightly different coordinate system

src/interpolation/polar_splines/polar_bsplines.hpp
EmilyBourne: This commit seems to do the opposite of what is requested

simulations/geometryRTheta/diocotron/CMakeLists.txt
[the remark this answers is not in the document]
PaulineVidal: I guess most of the operators can be applied (mathematically) in $(tor_1, tor_2)$ geometries. But they depends on the file `geometryRTheta`. I think maybe keep the suffix `*_RTheta` for these files. And once they will more general and do not depend on the geometry file `geometryRTheta`, we will call them `_polar`. What do you think?

src/interpolation/polar_splines/polar_bsplines.hpp
PaulineVidal: Sorry, I misunderstood. 
https://github.com/gyselax/gyselalibxx/pull/145/commits/c002729722d66afa1566e04fe3ec752bb6ba6156

simulations/geometryRTheta/diocotron/CMakeLists.txt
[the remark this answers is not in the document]
EmilyBourne: That sounds good to me :+1:

EmilyBourne (approved):
Looks good to me. Happy to merge when you feel ready to bring it out of draft

EmilyBourne:
> There is still the `Static Analyses` test which fails. I don't know what to do.

I only see 1 failing test. It is "CppCheck static analysis **warnings**" and is not marked as required

EmilyBourne:
Hmm, looking at the GPU results it seems that one of the variables is not as expected

PR #147, opened 2025-03-03
Check distribution function given to the collision operator (koliop) has power of two dimension extents

etiennemlb:
The operator expects powers of two for the dimensions. Enforce that constraint.


gdgirard (approved):
I have checked that it is OK from Gysela-Axi side. So fine with me.

PR #152, opened 2025-03-03
Correct the definition of norm_inf

### LINKED ISSUE #141: Suspicious definition of `norm_inf`
I think the scalar definition of the infinity norm should return the absolute value of the input scalar right ? 

https://github.com/gyselax/gyselalibxx/blob/0ba0292522eb7c5024efde412649c383eae7b69b/src/math_tools/l_norm_tools.hpp#L56-L70

EmilyBourne:
Fixes #141

### DIFF
--- a/src/math_tools/l_norm_tools.hpp
+++ b/src/math_tools/l_norm_tools.hpp
@@ -66,7 +66,7 @@ KOKKOS_FUNCTION double norm_inf(DVector<Tags...> vec)
  */
 KOKKOS_INLINE_FUNCTION double norm_inf(double const coord)
 {
-    return coord;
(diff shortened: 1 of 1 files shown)


PaulineVidal:
We didnt write any tests?

PaulineVidal:
We didnt write any tests?

PaulineVidal:
> > We didnt write any tests?
> > It doesnt seem to be the case. I dont know if it worth it to write one as we already did a mistake 😅
> 
> There doesn't seem to be any. I can add one if you think it is useful, I thought the test was likely to be so trivial that it was mostly useless

Probably true.

PaulineVidal (approved):

PR #162, opened 2025-03-05
Generalise interpolation rtheta

EmilyBourne:
Generalise interpolation rtheta to a 2D interpolation. This class is now independent of the R, Theta geometry.


src/interpolation/spline_interpolator_2d.hpp
```diff
@@ -0,0 +1,122 @@
...
+        return inout_data;
+    }
+};
+
+
+
+/**
+ * @brief A class which stores information necessary to create a pointer to an instance of the SplineInterpolator2D class.
+ *
+ * This class allows an instance of the SplineInterpolator2D class where necessary. This allows the
```
yanmnc, line 80: his class allows an instance of the SplineInterpolator2D class to be created where necessary. ?

EmilyBourne: Fixed 4cbe3c32

PR #164, opened 2025-03-05
Add a cache class for storing spline coefficients from a spline builder

yanmnc:
So that the builder may only be called once when computing partial derivatives of quantities (cf. another merge request to be added later on)


EmilyBourne (approved):
Looks good to me but please correct the warnings mentioned by the static analysis

PR #169, opened 2025-03-06
Reduce duplication in mapping tests

EmilyBourne:
Reduce duplication in mapping tests. This will reduce the number of changes in #127


tests/mapping/geometry_mapping_tests.hpp
```diff
@@ -0,0 +1,195 @@
...
+using IdxRangeRTheta = IdxRange<GridR, GridTheta>;
+
+
+template <class ElementType>
+using FieldMemRTheta_host = host_t<FieldMem<ElementType, IdxRangeRTheta>>;
+
+/**
+ * @brief Check if the product of the matrix and inv_matrix gives the identity matrix.
+ *
+ * The error tolerance is given at 1e-15.
```
yanmnc, line 147: 1e-10 ?

tests/mapping/geometry_mapping_tests.hpp
```diff
@@ -0,0 +1,195 @@
...
+                inv_tensor)
+{
+    double TOL = 1e-10;
+
+    using StartDim0 = ddc::type_seq_element_t<0, StartDims>;
+    using StartDim1 = ddc::type_seq_element_t<1, StartDims>;
+    using StartDim0_cov = typename StartDim0::Dual;
+    using StartDim1_cov = typename StartDim1::Dual;
+
+    using EndDim0_cov = ddc::type_seq_element_t<0, EndDims>;
```
yanmnc, line 167: Why is EndDims a typesea of covariant dimension and not contravariant like StartDims ?

tests/mapping/geometry_mapping_tests.hpp
```diff
@@ -0,0 +1,195 @@
...
+                            + ddcHelper::get<StartDim1, EndDim1_cov>(tensor)
+                                      * ddcHelper::get<EndDim1, StartDim0_cov>(inv_tensor);
+    EXPECT_NEAR(id_val10, 0., TOL);
+
+    double const id_val11 = ddcHelper::get<StartDim1, EndDim0_cov>(tensor)
+                                    * ddcHelper::get<EndDim0, StartDim1_cov>(inv_tensor)
+                            + ddcHelper::get<StartDim1, EndDim1_cov>(tensor)
+                                      * ddcHelper::get<EndDim1, StartDim1_cov>(inv_tensor);
+    EXPECT_NEAR(id_val11, 1., TOL);
+}
```
yanmnc, line 195: why don't you use the tensor_mul function ?

EmilyBourne: Fixed 661b1db0

tests/mapping/geometry_mapping_tests.hpp
EmilyBourne: No longer used after : 661b1db0

EmilyBourne: Originally I generalised this with the Jacobian in mind. The Jacobian is $J_i^j$ hence contravariant on the second index Ideally we shouldn't need to specify this though. The implementation is now more general

yanmnc (approved):

PR #174, opened 2025-03-07
Improve `tensor_mul` static assertions

### LINKED ISSUE #172: Improve tensor_mul assertions
The `tensor_mul` operator does not verify that the indices operator on the same dimensions. E.g.:
```cpp
DTensor<VectorIndexSet<X, Y>, VectorIndexSet<X, Y>> A;
DTensor<VectorIndexSet<R, Theta>> b;
tensor_mul(index<'i','j'>(A), index<'j'>(b));
```
should raise an error.

EmilyBourne:
Improve `tensor_mul` static assertions. Add the missing static assert to verify if the `VectorIndexSet`s that are indicated with the same character id are compatible. Fixes #172.
Add tests to ensure static assertions are correctly raised

### DIFF
--- a/src/data_types/indexed_tensor.hpp
+++ b/src/data_types/indexed_tensor.hpp
@@ -204,6 +204,11 @@ KOKKOS_FUNCTION auto tensor_mul(IndexedTensorType... tensor_to_mul)
             "You should not have more than two of any one index in an index expression. "
             "Additionally repeated indices should not be associated with two covariant or two "
             "contravariant indices.");
+    using SumVectorIndexIdMap = repeated_indices_t<AllIndexIdMaps>;
+    using SumUniqueIds = type_seq_unique_t<index_identifiers_t<SumVectorIndexIdMap>>;
+    static_assert(
+            ddc::type_seq_size_v<SumVectorIndexIdMap> == ddc::type_seq_size_v<SumUniqueIds>,
+            "Cannot sum over incompatible VectorIndexSets.");
     // Use the first tensor argument to extract the element type of the result.
     using ElementType = std::tuple_element_t<
             0,
@@ -212,7 +217,7 @@ KOKKOS_FUNCTION auto tensor_mul(IndexedTensorType... tensor_to_mul)
             (std::is_same_v<
                      typename IndexedTensorType::tensor_type::element_type,
                      ElementType> && ...),
-            "All tensors must have the same element type");
+            "All tensors in a tensor_mul must have the same element type.");
     using ResultIndexTypeSeq = non_repeated_indices_t<AllIndexIdMaps>;
     if constexpr (ddc::type_seq_size_v<ResultIndexTypeSeq> == 0) {
         ElementType result = 0.0;
--- a/src/data_types/indexed_tensor.hpp
+++ b/src/data_types/indexed_tensor.hpp
@@ -155,9 +155,9 @@ KOKKOS_FUNCTION void internal_tensor_mul(
 template <char ID, class AllIndexIdMaps>
 KOKKOS_INLINE_FUNCTION void check_id_validity()
 {
-    constexpr std::size_t n_occurences
+    constexpr std::size_t n_occurrences
             = char_occurrences_v<ID, index_identifiers_t<AllIndexIdMaps>>;
-    if constexpr (n_occurences == 2) {
+    if constexpr (n_occurrences == 2) {
         using RelevantVectorIndexSets = relevant_vector_index_sets_t<ID, AllIndexIdMaps>;
         using Set1 = ddc::type_seq_element_t<0, RelevantVectorIndexSets>;
         using Set2 = ddc::type_seq_element_t<1, RelevantVectorIndexSets>;
@@ -174,7 +174,7 @@ KOKKOS_INLINE_FUNCTION void check_id_validity()
                 "Cannot sum over incompatible VectorIndexSets.");
     } else {
         static_assert(
-                n_occurences == 1,
+                n_occurrences == 1,
                 "You should not have more than two of any one index in an index expression.");
     }
 }
--- a/src/data_types/indexed_tensor.hpp
+++ b/src/data_types/indexed_tensor.hpp
@@ -152,6 +152,39 @@ KOKKOS_FUNCTION void internal_tensor_mul(
     ((internal_tensor_mul_elem<GlobalTensorIndexIdMap, Is>(result, t...)), ...);
 }
 
+template <char ID, class AllIndexIdMaps>
+KOKKOS_INLINE_FUNCTION void check_id_validity()
+{
+    constexpr std::size_t n_occurences
+            = char_occurrences_v<ID, index_identifiers_t<AllIndexIdMaps>>;
+    if constexpr (n_occurences == 2) {
+        using RelevantVectorIndexSets = relevant_vector_index_sets_t<ID, AllIndexIdMaps>;
+        using Set1 = ddc::type_seq_element_t<0, RelevantVectorIndexSets>;
+        using Set2 = ddc::type_seq_element_t<1, RelevantVectorIndexSets>;
+        constexpr bool has_covariant_idx = (is_covariant_vector_index_set_v<Set1>)
+                                           or (is_covariant_vector_index_set_v<Set2>);
+        constexpr bool has_contravariant_idx = (is_contravariant_vector_index_set_v<Set1>)
+                                               or (is_contravariant_vector_index_set_v<Set2>);
+        static_assert(
+                has_covariant_idx and has_contravariant_idx,
+                "Repeated indices should not be associated with two covariant or two contravariant "
+                "indices.");
+        static_assert(
+                std::is_same_v<get_contravariant_dims_t<Set1>, get_contravariant_dims_t<Set2>>,
+                "Cannot sum over incompatible VectorIndexSets.");
+    } else {
+        static_assert(
+                n_occurences == 1,
+                "You should not have more than two of any one index in an index expression.");
+    }
+}
+
+template <class AllIndexIdMaps, class UniqueIds, std::size_t... Is>
+KOKKOS_INLINE_FUNCTION void check_all_id_validity(std::index_sequence<Is...>)
+{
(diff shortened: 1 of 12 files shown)


EmilyBourne:
This doesn't quite work as expected. I have an error:
> You should not have more than two of any one index in an index expression. Additionally repeated indices should not be associated with two covariant or two contravariant indices.

For:
```
/home/emily/Code/gyselalibxx/src/data_types/indexed_tensor.hpp: In instantiation of ‘auto tensor_mul(IndexedTensorType ...)
[with IndexedTensorType = {tensor_tools::IndexedTensor<
        const Tensor<double, ddc::detail::TypeSeq<X_pC, Y_pC>, ddc::detail::TypeSeq<X, Y> >,
        ddc::detail::TypeSeq<
               tensor_tools::VectorIndexIdMap<'i', ddc::detail::TypeSeq<X_pC, Y_pC> >,
               tensor_tools::VectorIndexIdMap<'j', ddc::detail::TypeSeq<X, Y> > > >,
    tensor_tools::IndexedTensor<
        const Tensor<double, ddc::detail::TypeSeq<X, Y> >,
        ddc::detail::TypeSeq<
               tensor_tools::VectorIndexIdMap<'j', ddc::detail::TypeSeq<X, Y> > > >}]’:
```
This is because `X_cov == X`

tests/data_types/static_assert_tests/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+function(add_static_assert_test file error_msg)
+
+add_executable(${file}-build-fail-test
+    ${file}.cpp
+    ../../main.cpp
+)
+target_link_libraries(${file}-build-fail-test
+        PUBLIC
+        GTest::gtest
+        GTest::gmock
```
tpadioleau, line 11: Can we get rid off google test for compilation-only tests ?

src/data_types/indexed_tensor.hpp
```diff
@@ -152,6 +152,39 @@ KOKKOS_FUNCTION void internal_tensor_mul(
...
+        using Set1 = ddc::type_seq_element_t<0, RelevantVectorIndexSets>;
+        using Set2 = ddc::type_seq_element_t<1, RelevantVectorIndexSets>;
+        constexpr bool has_covariant_idx = (is_covariant_vector_index_set_v<Set1>)
+                                           or (is_covariant_vector_index_set_v<Set2>);
+        constexpr bool has_contravariant_idx = (is_contravariant_vector_index_set_v<Set1>)
+                                               or (is_contravariant_vector_index_set_v<Set2>);
+        static_assert(
+                has_covariant_idx and has_contravariant_idx,
+                "Repeated indices should not be associated with two covariant or two contravariant "
+                "indices.");
```
tpadioleau, line 171: I think we should avoid mixing notations for operators like `&&`/`and`, `||`/`or` etc

tests/data_types/static_assert_tests/CMakeLists.txt
EmilyBourne: Apparently yes : f397e05f

src/data_types/indexed_tensor.hpp
EmilyBourne: Fixed  ce2738e

EmilyBourne:
> I am a bit confused between IdMaps and VectorIndexSets and RelevantIndexIndexSets. I will try to review again on Monday. I left a few remarks on the form.

As a reminder:
https://github.com/gyselax/gyselalibxx/blob/f46d07a142275e5ada6d183a1f667f500fe3e527/src/data_types/vector_index_tools.hpp#L9-L10

https://github.com/gyselax/gyselalibxx/blob/f46d07a142275e5ada6d183a1f667f500fe3e527/src/data_types/vector_index_tools.hpp#L125-L135

The `RelevantVectorIndexSets` for a character `c` are all the `VectorIndexSets` which are found in a `VectorIndexIdMap` whose `id` is equal to the character `c`.

EmilyBourne:
> I am a bit confused between IdMaps and VectorIndexSets and RelevantIndexIndexSets. I will try to review again on Monday. I left a few remarks on the form.

@tpadioleau I agree that the are several objects in this part of the code with very similar names. If you have any suggestions of clearer names to help disambiguate then this would be appreciated

tests/data_types/static_assert_tests/CMakeLists.txt
tpadioleau: I mean, `add_static_assert_test` should not even need googletest as it does not use any runtime `ASSERT`/`EXPECT` right ? I think if we also remove `GTest::gmock` (that actually brings transitively `GTest::gtest`) we should be fine. The file `main.cpp` will complain but we don't need it either.

EmilyBourne: I think you need at least one for the `#include <gtest/gtest.h>` to be correctly interpreted.

tpadioleau: What about writing the tests like this:
```cpp
using Tensor2D_A = Tensor<int, VectorIndexSet<R, Theta>, VectorIndexSet<R_cov, Theta_cov>>;
using Tensor2D_B = Tensor<int, VectorIndexSet<X, Y>, VectorIndexSet<R_cov, Theta_cov>>;
using Tensor2D_C = Tensor<int, VectorIndexSet<R, Theta>, VectorIndexSet<R_cov, Theta_cov>>;

Tensor2D_C compile_tensor_test_mul(Tensor2D_A& A, Tensor2D_B& B)
{
    ddcHelper::get<R, R>(A) = 1;
    ddcHelper::get<R, Theta>(A) = 0;
    ddcHelper::get<Theta, R>(A) = 2;
    ddcHelper::get<Theta, Theta>(A) = 4;
    ddcHelper::get<X, R_cov>(B) = 6;
    ddcHelper::get<X, Theta_cov>(B) = 8;
    ddcHelper::get<Y, R_cov>(B) = 4;
    ddcHelper::get<Y, Theta_cov>(B) = 5;
    return tensor_mul(index<'i', 'j'>(A), index<'j', 'k'>(B));
}
```

EmilyBourne: Ah ok. Why not, maybe it will make the compilation faster :+1:

EmilyBourne: Done : 100615b11fab83dfea2bc81e2f2d37fe0d1485e5

src/data_types/indexed_tensor.hpp
```diff
@@ -152,6 +152,39 @@ KOKKOS_FUNCTION void internal_tensor_mul(
...
+    constexpr std::size_t n_occurrences
+            = char_occurrences_v<ID, index_identifiers_t<AllIndexIdMaps>>;
+    if constexpr (n_occurrences == 2) {
+        using RelevantVectorIndexSets = relevant_vector_index_sets_t<ID, AllIndexIdMaps>;
+        using Set1 = ddc::type_seq_element_t<0, RelevantVectorIndexSets>;
+        using Set2 = ddc::type_seq_element_t<1, RelevantVectorIndexSets>;
+        constexpr bool has_covariant_idx = (is_covariant_vector_index_set_v<Set1>)
+                                           || (is_covariant_vector_index_set_v<Set2>);
+        constexpr bool has_contravariant_idx = (is_contravariant_vector_index_set_v<Set1>)
+                                               || (is_contravariant_vector_index_set_v<Set2>);
```
tpadioleau, line 167: I hope my question makes sense, is the whole vector index set supposed to be fully covariant or contravariant ? The vector set cannot have a mix of covariant/contravariant indices ?

tests/data_types/static_assert_tests/CMakeLists.txt
tpadioleau: We should think of making this snippet of cmake code available for other projects. I don't know if it is robust enough to capture correctly the message in the `static_assert` with other compilers than the ones in the CI ?

src/data_types/indexed_tensor.hpp
EmilyBourne: The question makes sense. While it is theoretically possible to build a vector containing both contravariant and covariant components it makes no sense to do so. The covariant and contravariant indices represent different bases. So writing `DVector<R, Theta_cov>` makes as much sense as `DVector<R, X>`

tpadioleau: mmh are you saying that $a_i^j = b_{i, k} c^{k, j}$ does not make sense ?

EmilyBourne: No. This is ok. I do this for the Jacobian. It means that the matrix is expressed using 2 vector index sets.
E.g 
$J_i^k=\frac{\partial \xi_i}{\partial \eta_j}$
With $\xi_i$ coming from the VectorIndexSet {x,y} and $\eta_k$ coming from the VectorIndexSet {r,theta} (derivatives lead to inverted indices this is actually $`{\partial_\eta}^k`$ from the VectorIndexSet $`\{\partial^r, \partial^\theta\}`$ in tensor notation)

EmilyBourne: The row indices are associated with one basis while the column indices are associated with another. What does not make sense is having row indices that represent elements from a variety of bases

EmilyBourne: $a_i^j$ should have a type like:
```cpp
DTensor<VectorIndexSet<R, Theta>, VectorIndexSet<R_cov, Theta_cov>>
```
Each VectorIndexSet is either covariant or contravariant but the matrix has one index from each set

tpadioleau: Oh I see thanks

tpadioleau (approved):
Ok for me

tests/data_types/static_assert_tests/CMakeLists.txt
EmilyBourne: I found the instructions on the internet so I'm not sure it's mine to share. I hope it is robust enough as it just checks that the static assert message is raised. That should always be the case unless the assertion is broken I think?

PR #178, opened 2025-03-10
Fix definition of the advection field on RTheta

PaulineVidal:
Fix the definition of the advection field on the (r,theta) axis. 

The advection field is now expressed in the contravariant basis.


yanmnc:
https://github.com/gyselax/gyselalibxx/blob/ba0c1f98cf697925babe7b93696f529096eacdaa/src/geometryRTheta/advection_field/README.md?plain=1#L152

To denote a component of the metric tensor you can write directly 

$`G = J^T J `$ the metric tensor, whose components are $`g_{ij}`$

yanmnc:
https://github.com/gyselax/gyselalibxx/blob/ba0c1f98cf697925babe7b93696f529096eacdaa/src/geometryRTheta/advection_field/README.md?plain=1#L153

By default $`g^{ij}`$ **is** the ij component of $`G^{-1}`$. I would suggest to write 

$`G^{-1}`$ the inverse metric tensor,  whose components are $`g^{ij}`$

src/geometryRTheta/advection_field/README.md
```diff
@@ -155,7 +155,7 @@ with
 
 In 2D, it can be rewritten as the following matrix system 
 ```math
-\nabla f = 
+\hat{\nabla} f =
```
yanmnc, line 158: Tbh I'm a bit confused by this notation, I understand that it results from the following idea: whenever a vector is written with a hat, it means that when writing its components the pseudocartesian basis is used. As mentioned in my previous comment, a vector quantity is the same physical object independently of the chosen basis, so `$\hat{\nabla} f$` is really the same thing as `$\nabla f$`. What do you think of dropping the hat notations for vectors, but keeping it for their components ? I think it would be more rigorous. For instance you could write 

```math
\nabla f =
\begin{bmatrix}
    (\hat{\nabla} f)_r \\
   (\hat{\nabla} f)_{\theta}\\
\end{bmatrix} =
\begin{bmatrix}
    \hat{\nabla}_r f \\
   \hat{\nabla}_{\theta} f\\
\end{bmatrix} =
G^{-1}
\begin{bmatrix}
    \partial_{x_1} f \\
    \partial_{x_2} f \\
\end{bmatrix}
```
```

File: src/geometryRTheta/advection_field/README.md [hunk 2]
```diff
@@ -215,24 +215,34 @@ E
...
-A
-= E \wedge e_z
+\hat{A}_{r,\theta} = J^{-1} A_{x,y} = J^{-1}(E_{x,y} \wedge e_z) = J^{-1}(J \hat{E}_{r,\theta} \wedge e_z)
 = 
+\frac{1}{\det(J)}
 \begin{bmatrix}
-    -E_{\theta} \\
-    E_r \\
-\end{bmatrix}.
+    -(J_{1,1}J_{1,2} + J_{2,1}J_{2,2})  & -(J_{2,2}^2 + J_{1,2}^2) \\
```
yanmnc, line 222: There should be no minus sign here I think, rather 

```math
\begin{bmatrix}
    J_{1,1}J_{1,2} + J_{2,1}J_{2,2}  & J_{2,2}^2 + J_{1,2}^2 \\
     -(J_{1,1}^2 + J_{2,1}^2)            &  -(J_{1,1}J_{1,2} + J_{2,1}J_{2,2}) \\
\end{bmatrix}
```

yanmnc (changes requested):

src/geometryRTheta/advection_field/README.md
PaulineVidal: It seems to be something applied. I follow the rules on this article
https://www.sciencedirect.com/science/article/pii/S002199911930587X
image

But I can remove them if you prefer, I just thought it coulb be clearer.

src/geometryRTheta/advection/bsl_advection_rtheta.hpp
```diff
@@ -181,24 +181,19 @@ class BslAdvectionRTheta : public IAdvectionRTheta
...
+            std::array<std::array<double, 2>, 2> J; // Jacobian matrix
+            m_mapping.jacobian_matrix(coord_rtheta, J);
 
             ddcHelper::get<X>(advection_field_xy_host)(irtheta)
-                    = ddcHelper::get<R_cov>(advection_field_rtheta)(irtheta) * inv_J[0][0]
-                              * jacobian
-                      + ddcHelper::get<Theta_cov>(advection_field_rtheta)(irtheta) * inv_J[1][0]
-                                * jacobian;
+                    = ddcHelper::get<R_cov>(advection_field_rtheta)(irtheta) * J[0][0]
+                      + ddcHelper::get<Theta_cov>(advection_field_rtheta)(irtheta) * J[0][1];
```
yanmnc, line 193: I think that because of [this discussion](https://gysela.slack.com/archives/C08H1DF0HU2/p1741622750321159), the present formula is now wrong O:)

yanmnc (changes requested):

tests/geometryRTheta/advection_field_rtheta/advection_field_gtest.cpp
```diff
@@ -250,19 +250,16 @@ TEST(AdvectionFieldRThetaComputation, TestAdvectionFieldFinder)
...
+        std::array<std::array<double, 2>, 2> J; // Jacobian matrix
+        to_physical_mapping.jacobian_matrix(coord_rtheta, J);
 
         // computation made in BslAdvectionRTheta operator:
         ddcHelper::get<X>(advection_field_xy_from_rtheta)(irtheta)
-                = ddcHelper::get<R_cov>(advection_field_rtheta)(irtheta) * inv_J[0][0] * jacobian
-                  + ddcHelper::get<Theta_cov>(advection_field_rtheta)(irtheta) * inv_J[1][0]
-                            * jacobian;
+                = ddcHelper::get<R_cov>(advection_field_rtheta)(irtheta) * J[0][0]
+                  + ddcHelper::get<Theta_cov>(advection_field_rtheta)(irtheta) * J[0][1];
```
yanmnc, line 259: same remark as [here](https://github.com/gyselax/gyselalibxx/pull/178#discussion_r1987647925)

PaulineVidal:
I start to doubt of everything. 

1. Do we agree that for the electric field the gradient in the formula is on the _covariant_ vectors? 

I mean, 
```math 
E^{xy} = - \nabla^{xy} \phi
```

and it would be 
```math 
E^{r\theta} = - \nabla^{r\theta} \phi = - G^{-1} d_{r\theta} \phi
```

2. And if I express  the advection equation in the logical domain, it is more natural to use the _covariant_ vectors or the _contravariant_ vectors? 
```math 
(1)  \quad \partial_t f + A^{r\theta?}_{r\theta?}\cdot \nabla^{r\theta} f = 0, 
```

or 
```math 
(2) \quad \partial_t f + A^{r\theta?}_{r\theta?}\cdot d_{r\theta} f = 0.
```

For the advection field, I would be tempted to say that in (1), it should be $A^{r\theta}\cdot \nabla^{r\theta} f$ and 
in (2), $A_{r\theta}\cdot d_{r\theta} f $. But, I am clearly not sure. 

Well, I guess it is a matter of choice.. So the question is more what is the most natural?

yanmnc:
@PaulineVidal concerning [your question](https://github.com/gyselax/gyselalibxx/pull/178#issuecomment-2711509397): 

1. You can express the electric field (a vector) by writing its decomposition on either the contravariant bases $`\{b_i\}`$ or the covariant bases $`\{b^i\}`$: 
```math
E = E^i b_i = E_i b^i
``` 

For instance: $`E = E^r e_r + E^{\theta}e_{\theta}`$

2. Both are possible, the "natural" basis is the contravariant one (it is the one you encounter usually in physics textbook, like the cylindrical contravariant basis $`(e_r, e_{\theta}, e_z)`$. 

In any case the following holds 
```math
\overrightarrow{A} \cdot \overrightarrow{\nabla} f  = (A_i b^i) \cdot (\nabla^j f \, b_j) = (A^i b_i) \cdot (\frac{\partial f}{\partial q^j} b^j) 
``` 

With 

```math
\nabla^j f  = \frac{\partial f}{\partial q^i} g^{ij}
``` 

and $`\{q^i\}`$ is a set of curvilinear coordinates.

src/geometryRTheta/advection_field/README.md
```diff
@@ -189,23 +189,23 @@ J^{-T}
...
 with $`\nabla_{y_1, y_2} f = [\partial_{y_1} f, \partial_{y_2} f]^T`$ and 
-$`\nabla_{x_1, x_2} f = \sum_i \sum_j \partial_{x_i} f g^{ij} e_j`$.
+$` \hat{\nabla}_{x_1, x_2} f = \sum_i \sum_j \partial_{x_i} f g^{ij} e_j`$.
 
 
 #### Application to the advection field
 * In our case, we use this formula to compute the electric field along the logical axis: 
 ```math
-E
+\hat{E}
```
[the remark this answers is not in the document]
PaulineVidal: I removed the hat.

src/geometryRTheta/advection_field/README.md
PaulineVidal: I desagree. Why ? 
And it seems to work as I did :thinking: 
Maybe this comment is not anymore updated.

src/geometryRTheta/advection_field/README.md
```diff
@@ -141,90 +124,134 @@ A = E\wedge e_z
...
+with $`G^{-1}`$ the inverse [metric tensor](#docs_mathematical_and_physical_coneventions__Metric_tensor). 
+
+Then the advection field is given by 
 ```math
-\nabla_{y_1, y_2} f = J \nabla_{x_1, x_2} f,
+A = E \wedge e_z = 
+\begin{bmatrix}
+    - E \cdot e_x \\
+      E \cdot e_y \\
+\end{bmatrix},
```
EmilyBourne, line 188: Are you sure about this? Shouldn't it be:
```suggestion
A = E \wedge e_z = 
\begin{bmatrix}
      E \cdot e_y \\
    - E \cdot e_x \\
\end{bmatrix},
```

src/geometryRTheta/advection_field/README.md
EmilyBourne: This comment is indeed outdated (see the header):
image

However the final expression has not changed. I think you have simply moved the error to here:
https://github.com/gyselax/gyselalibxx/pull/178/files#r1992985435

> And it seems to work as I did

Are you sure? Isn't this version virtually untested as it is the version in $(r,\theta)$ coordinates that you provided for Virginie but do not use in your simulations?

I think that the only test copy/pastes this expression so it does not check its validity

src/geometryRTheta/advection_field/README.md
PaulineVidal: Fixed. 
[fix wedge operator](https://github.com/gyselax/gyselalibxx/pull/178/commits/c240e2b321f7d1eb7f907b23a6f158478fa6ef10)
[Fix definition of the electric potential to match with the fixed defi…](https://github.com/gyselax/gyselalibxx/pull/178/commits/f1f154562c29b6e8a1a381962ff098cd42927d50)
[Change the contravariant basis to covariant basis to reduce the compu](https://github.com/gyselax/gyselalibxx/pull/178/commits/7c7915c52d28f087fe669cdf6595b5d77d8d7963)

tests/geometryRTheta/advection_field_rtheta/advection_field_gtest.cpp
PaulineVidal: Fixed ["characteristic"](https://github.com/gyselax/gyselalibxx/pull/178/commits/6543d6753511a0900554393630efcffa3eeeec4a)

src/geometryRTheta/advection/bsl_advection_rtheta.hpp
```diff
@@ -17,38 +17,38 @@
...
- * @f$ \partial_t X(t; s, x) = V(t, X(t; s, x)), \qquad \text{ with } X(s; s, x) = x. @f$
+ * @f$ \partial_t X(t; s, x) = A(t, X(t; s, x)), \qquad \text{ with } X(s; s, x) = x. @f$
  *
  * Then the property gives us:
  * @f$ f(t, x) = f(0, X(t; 0, x)), \quad \forall t. @f$
  *
  *
- * So the first step of the advection operator is to compute the characteristic feet @f$ X(t; t+dt, x_i) @f$
- * for each mesh point @f$ x_i @f$.
+ * So the first step of the advection operator is to compute the feet of the characteristic
```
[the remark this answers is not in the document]
PaulineVidal: ["characteristic"](https://github.com/gyselax/gyselalibxx/pull/178/commits/6543d6753511a0900554393630efcffa3eeeec4a)

src/geometryRTheta/advection_field/README.md
```diff
@@ -106,31 +98,25 @@ From these equations, we deduce the (unique) values of $`\partial_x\phi`$ and $`
...
+    \partial_r \phi (0, \theta_2) \\
 \end{bmatrix}.
 ```
 
 Then we compute $`E`$ at $`(x,y) = (0,0)`$ and $`(x,y) = \mathcal{F}(\varepsilon,\theta)`$ $`\forall \theta`$ (for $`\varepsilon\neq 0`$, we use the Jacobian matrix as previously) and we linearise
 
 ```math
-E_x(r, \theta) = \left( 1 - \frac{r}{\varepsilon} \right)  E_x(0, \theta) + \frac{r}{\varepsilon} E_x(\varepsilon, \theta), \\
-E_y(r, \theta) = \left( 1 - \frac{r}{\varepsilon} \right)  E_y(0, \theta) + \frac{r}{\varepsilon} E_y(\varepsilon, \theta), 
+E(r, \theta) = \left( 1 - \frac{r}{\varepsilon} \right)  E(0, \theta) + \frac{r}{\varepsilon} E(\varepsilon, \theta), \\
```
yanmnc, line 109:
```suggestion
E(r, \theta) = \left( 1 - \frac{r}{\varepsilon} \right)  E(0, \theta) + \frac{r}{\varepsilon} E(\varepsilon, \theta),
```

src/geometryRTheta/advection/bsl_advection_rtheta.hpp
```diff
@@ -156,6 +156,7 @@ class BslAdvectionRTheta : public IAdvectionRTheta
      * @param [in] advection_field_rtheta
      *      A DConstVectorFieldRTheta containing the values of the advection field
      *      on the logical index range axis.
+     *      It is expressed in the contravariant basis.
```
EmilyBourne, line 159:
```suggestion
     *      It is expressed on the contravariant basis.
```

src/geometryRTheta/advection/iadvection_rtheta.hpp
```diff
@@ -39,7 +39,7 @@ class IAdvectionRTheta
      * @param[in, out] allfdistribu
      *      The function to be advected.
      * @param[in] advection_field
-     *      The advection field along the logical index range axes, RTheta.
+     *      The advection field in the contravariant basis of the logical domain.
```
EmilyBourne, line 42:
```suggestion
     *      The advection field on the contravariant basis of the logical domain.
```

src/geometryRTheta/advection/README.md
```diff
@@ -45,9 +45,9 @@ There are multiple time integration methods available which are implemented in t
 
  
  
-We are listing the different schemes for this equation $`\partial_t X (t) = A_x(t, X(t),Y(t))`$. 
+We are listing the different schemes for this equation $`\partial_t X (t) = A(t, X(t))`$. 
 
-We write $X (t) = X (t; s, x, y)$, $X^n = X(t^n)$ and $A^n(X) = A(t^n, X)$ for a time discretisation 
+We write $X (t) = X (t; s, x)$, $X^n = X(t^n)$ and $A^n(X) = A(t^n, X)$ for a time discretisation
```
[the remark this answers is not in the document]
PaulineVidal: To make it fit with the definition of A. Here I am solving an ODE. As I am writting only the equations on X, I think I can introduce a 1D equation.

src/geometryRTheta/advection_field/README.md
```diff
@@ -26,8 +26,8 @@ with $`\rho`$ the density, $`\phi`$ the electrostatic potential and $`E`$ the el
 
 The AdvectionFieldFinder computes the advection field $`A`$ from the electrical field $`\phi`$ returned by the PolarSplineFEMPoissonLikeSolver. 
 It has two types of `operator()`: 
-* one returning the advection field along the axis of the physical domain: $`A = (A_x, A_y)`$
-* and the another returning the advection field along the axis of the logical domain: $`A = (A_r, A_\theta)`$. 
+* one returning the advection field expressed on the vectors $`(e_x, e_y)`$ of the physical domain: $`A = A_x e_x + A_y e_y`$
+* and the another returning the advection field expressed on the vectors $`(e^r, e^\theta)`$ of the [contravariant basis](#docs_mathematical_and_physical_conventions) on the logical domain: $`A = A^r e_r + A^\theta e_\theta`$.
```
EmilyBourne, line 30:
```suggestion
* and the other returning the advection field expressed on the vectors $`(e^r, e^\theta)`$ of the [contravariant basis](#docs_mathematical_and_physical_conventions) of the logical domain: $`A = A^r e_r + A^\theta e_\theta`$. 
```
or
```suggestion
* and another returning the advection field expressed on the vectors $`(e^r, e^\theta)`$ of the [contravariant basis](#docs_mathematical_and_physical_conventions) of the logical domain: $`A = A^r e_r + A^\theta e_\theta`$. 
```

src/geometryRTheta/advection_field/README.md
```diff
@@ -50,40 +50,32 @@ The computation of the electrical field can be ill-defined around the O-point, s
...
-\end{bmatrix}
+\nabla \phi
+= \partial_x \phi e_x + \partial_y \phi e_y 
+= J^{-T}(\partial_r \phi e^r + \partial_\theta \phi e^\theta), 
+\qquad 
+\text{in the covavariant basis,}
 ```
 
-with $`J`$  the Jacobian matrix of the mapping $`\mathcal{F}: (r,\theta)\mapsto(x,y)`$. Then the electric field is given by 
+with $`J`$  the Jacobian matrix and *G* the tensor metric of the mapping $`\mathcal{F}: (r,\theta)\mapsto(x,y)`$.
```
EmilyBourne, line 60: > *G* the tensor metric of the mapping $`\mathcal{F}: (r,\theta)\mapsto(x,y)`$.

I am not a fan of this definition. *G* is associated with a basis not a mapping.

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
```diff
@@ -23,27 +23,27 @@
  *
  * The Vlasov-Poisson equations are given by
  *
- * - (1) @f$ \partial_t \rho - E_y \partial_x \rho + E_x \partial_y\rho = 0 @f$,
+ * - (1) @f$ \partial_t \rho + (E \wedge e_z) \cdot \nabla \rho = 0 @f$,
  *
- * - (2) @f$ - L \phi = - \nabla \cdot (\alpha \nabla \phi) + \beta \phi = \rho  @f$,
+ * - (2) @f$ - \nabla \phi = \rho  @f$,
```
EmilyBourne, line 28: Are you sure about this change? You now have $E = \rho$

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
```diff
@@ -252,17 +253,20 @@ class AdvectionFieldFinder
...
-                Matrix_2x2 inv_J = inv_jacobian_matrix(coord_rtheta);
+                Matrix_2x2 inv_J = inv_jacobian_matrix(coord_rtheta); // inverse Jacobian matrix
 
-                // Gradient of phi in the physical index range (Cartesian index range)
-                double const deriv_x_phi = deriv_r_phi(irtheta) * inv_J[0][0]
-                                           + deriv_theta_phi(irtheta) * inv_J[1][0];
-                double const deriv_y_phi = deriv_r_phi(irtheta) * inv_J[0][1]
-                                           + deriv_theta_phi(irtheta) * inv_J[1][1];
+                // Gradient of phi in the physical domain (Cartesian domain)
+                // grad phi = (dr phi, dtheta phi) in coravariant basis
```
EmilyBourne, line 259:
```suggestion
                // grad phi = (dr phi, dtheta phi) in covariant basis
```

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
```diff
@@ -302,7 +306,8 @@ class AdvectionFieldFinder
                 // --- Value at r = m_epsilon:
                 CoordRTheta const coord_rtheta_epsilon(m_epsilon, th);
 
-                Matrix_2x2 inv_J_eps = inv_jacobian_matrix(coord_rtheta_epsilon);
+                Matrix_2x2 inv_J_eps
+                        = inv_jacobian_matrix(coord_rtheta_epsilon); // inverse Jacobian matrix
```
EmilyBourne, line 310: Do you think this comment is necessary? Could it be avoided by simply renaming `inv_jacobian_matrix` to `inverse_jacobian_matrix`?

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
```diff
@@ -311,15 +316,20 @@ class AdvectionFieldFinder
...
 
-                // Gradient of phi in the physical index range (Cartesian index range)
-                double const deriv_x_phi_epsilon = deriv_r_phi_epsilon * inv_J_eps[0][0]
-                                                   + deriv_theta_phi_epsilon * inv_J_eps[1][0];
-                double const deriv_y_phi_epsilon = deriv_r_phi_epsilon * inv_J_eps[0][1]
-                                                   + deriv_theta_phi_epsilon * inv_J_eps[1][1];
+                // Gradient of phi in the physical domain (Cartesian domain)
+                // grad phi = (dr phi, dtheta phi) in covariant basis
+                double const grad_r_phi_epsilon = deriv_r_phi_epsilon;
+                double const grad_theta_phi_epsilon = deriv_theta_phi_epsilon;
```
EmilyBourne, line 322: Why introduce an additional variable rather than just renaming `deriv_r_phi_epsilon` and `deriv_theta_phi_epsilon`?

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
```diff
@@ -484,7 +495,10 @@ class AdvectionFieldFinder
             CoordRTheta const coord_rtheta(ddc::coordinate(irtheta));
 
             DTensor<VectorIndexSet<R_cov, Theta_cov>, VectorIndexSet<R_cov, Theta_cov>> inv_G
-                    = metric_tensor.inverse(coord_rtheta);
+                    = metric_tensor.inverse(coord_rtheta); // inverse tensor metric
```
EmilyBourne, line 498: Similar question to above. Is it really that hard to recognise that `metric_tensor.inverse` is the inverse tensor metric (or rather the inverse metric tensor)?

EmilyBourne (changes requested):
This PR is now nearly 400 lines long. It seems that several of these changes are unrelated to the subject of the PR. I would prefer that we split this PR in two to allow us to concentrate on the equations while reviewing

PaulineVidal:
> This PR is now nearly 400 lines long. It seems that several of these changes are unrelated to the subject of the PR. I would prefer that we split this PR in two to allow us to concentrate on the equations while reviewing

Arround 20 lines changed correspond to the change "index range" -> "domain", "characteristics feet" -> "feet of the characteristic". Does it really need to be split? Otherwise, I think it match with "fix the advection field" :sweat_smile:

src/geometryRTheta/advection_field/README.md
PaulineVidal: According to the note, the metric tensor corresponds to the scalar product of the vectors of the contravariant basis of the mapping. I thought that once the mapping determined, the definition of the metric tensor is fixed.

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
PaulineVidal: Is it a problem if there are these comments? Orginally the comments were because it might be not immediate that J or inv_G refers to the Jacobian _matrix_ or the inverse _metric tensor_ but convenient to work with small variable name in the lines below.

src/geometryRTheta/advection_field/README.md
EmilyBourne: Not quite. The metric tensor corresponds to the scalar product of the vectors of a contravariant basis.

So for example you have the metric tensor of the cartesian basis $G_c=e_i\cdot e_j$ which is the identity, the metric tensor of the curvilinear basis $G=b_i\cdot b_j$, and the metric tensor of the pseudo-cartesian basis $G=pc_i\cdot pc_j$

The word "mapping" does not appear in the note because there can be multiple mappings in a simulation and there are bases associated with both the departure space and the arrival space. Therefore while the "mapping" can be used as a tool to calculate the metric tensor associated with a basis, it is not itself sufficiently precise to define the metric tensor

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
EmilyBourne: > Is it a problem if there are these comments?

No it is not a problem. It is just an extra thing to maintain. Given that there wasn't a comment here previously and you have added one I assume that you feel that it adds extra information to the line. I was wondering what as for me this is already clear

EmilyBourne: Generally comments are good when they clarify the code and bad when they repeat the code. E.g. the code would be much denser if we had lines like:
```
c = a + b; // Add a and b and save in c
```

PaulineVidal:
I put the modifications in #189 that we can merge before .

src/geometryRTheta/advection/README.md
```diff
@@ -21,8 +21,8 @@ It uses the conservation property along the characteristics:
 
 with:
 ```math
-\partial_t X (t; s, x, y) = A_x(t,X(t; s, x, y),Y(t; s, x, y)),\\
-\partial_t Y (t; s, x, y) = A_y(t,X(t; s, x, y),Y(t; s, x, y)),\\
+\partial_t X (t; s, x, y) = A(t,X(t; s, x, y),Y(t; s, x, y)) \cdot e_x,\\
+\partial_t Y (t; s, x, y) = A(t,X(t; s, x, y),Y(t; s, x, y)) \cdot e_y,\\
 X(s; s, x, y) = x,\\
 Y(s; s, x, y) = y.
```
yanmnc, line 27:
```suggestion
\begin{aligned}
\partial_t X (t; s, x, y) = A(t,X(t; s, x, y),Y(t; s, x, y)) \cdot e_x,\\
\partial_t Y (t; s, x, y) = A(t,X(t; s, x, y),Y(t; s, x, y)) \cdot e_y,\\
X(s; s, x, y) = x,\\
Y(s; s, x, y) = y.
\end{aligned}
```

src/geometryRTheta/advection_field/README.md
```diff
@@ -141,98 +127,142 @@ A = E\wedge e_z
...
 ```
 
 with 
-* $`J`$ the Jacobian matrix associated with the mapping function of the system $`\mathcal{F}:(x_1, x_2)\mapsto(y_1,y_2)`$, 
-* $`G = J^T J = [g_{ij}]_{ij}`$ the metric tensor, 
-* $`G^{-1} = [g^{ij}]_{ij}`$ the inverse metric tensor 
-* and $`e_j`$ the unnormalised local covariant vectors. 
+* $`J`$ the Jacobian matrix associated with the mapping function of the system $`\mathcal{F}:(q_1,..., q_N)\mapsto(x_1, ..., x_N)`$, 
+* $`G = J^T J`$ the metric tensor, whose components are $`g_{ij}`$,  
+* $`G^{-1}`$ the inverse metric tensor, whose components are $`g^{ij}`$,
```
yanmnc, line 139: these quantities do not appear in the above equation

EmilyBourne (approved):

PR #185, opened 2025-03-14
Use allfdistribu_predicted in bsl_predcorr_second_order_implicit.hpp

### LINKED ISSUE #139: `allfdistribu_predicted` is unused in bsl_predcorr_second_order_implicit.hpp
`allfdistribu_predicted` is calculated here:
https://github.com/gyselax/gyselalibxx/blob/2d546a9a0f3c1cddc7739b5c9f278b40aa311627/src/geometryRTheta/time_solver/bsl_predcorr_second_order_implicit.hpp#L278-L283

however once calculated it is never used. It seems that this variable can therefore be removed but I suspect that this is not the intention of the equations. Probably it should be used here:
https://github.com/gyselax/gyselalibxx/blob/2d546a9a0f3c1cddc7739b5c9f278b40aa311627/src/geometryRTheta/time_solver/bsl_predcorr_second_order_implicit.hpp#L292-L296

@PaulineVidal can you confirm?

PaulineVidal:
To fix #139


EmilyBourne (approved):

src/geometryRTheta/time_solver/bsl_predcorr_second_order_implicit.hpp
```diff
@@ -297,7 +297,7 @@ class BslImplicitPredCorrRTheta : public ITimeSolverRTheta
 
 
             // STEP 4: From rho^P, we compute phi^P: Poisson equation
-            m_builder(get_field(allfdistribu_coef), get_const_field(allfdistribu_host));
+            m_builder(get_field(allfdistribu_coef), get_const_field(allfdistribu_predicted));
```
EmilyBourne, line 300: There seems to be a problem here because `allfdistribu_predicted` is on the device while `allfdistribu_host` was on the host. Maybe a missing `ddc::create_mirror_view_and_copy`

```suggestion
            auto allfdistribu_predicted_host = ddc::create_mirror_view_and_copy(allfdistribu_predicted);
            m_builder(get_field(allfdistribu_coef), get_const_field(allfdistribu_predicted_host));
```

EmilyBourne:
To merge when https://gitlab.maisondelasimulation.fr/gysela-developpers/gyselalibxx/-/jobs/170568 passes

EmilyBourne (approved):

PR #187, opened 2025-03-17
Use GPU-direct MPI

EmilyBourne:


EmilyBourne:
@tpadioleau does this PR require something to be updated in persee's setup?

tpadioleau:
Should be ok to run on Persee we rely on
```
openmpi +cuda
```
I think we should be careful of synchronization, I think the previous implementation was relying on the host-device synchronization from `ddc::create_mirror_view_and_copy`.

Isn't it a bit unfortunate to loose the ability to run on non GPU-aware MPI implementations ? AFAIK it is an extension of the MPI specification, implementations do not have to implement this feature for all GPU vendors.

EmilyBourne:
> I think we should be careful of synchronization, I think the previous implementation was relying on the host-device synchronization from `ddc::create_mirror_view_and_copy`.

I think so. But if the MPI command now runs on the GPU do we need a synchronisation?

> Isn't it a bit unfortunate to loose the ability to run on non GPU-aware MPI implementations ? AFAIK it is an extension of the MPI specification, implementations do not have to implement this feature for all GPU vendors.

We don't 100% lose the ability as the `operator()` of `MPITransposeAllToAll` takes an `ExecSpace` so you can request CPU. But in this case the pre and post transposes will also be done on CPU

EmilyBourne:
> unlike nccl or rccl, the mpi std does guarantee when the data is ready (you can expect it to be in sync with the kernels on the null stream) you habe to mpi_wait (or similar)

@etiennemlb should this read "doesn't guarantee when the data is ready" ?

tpadioleau:
> Use GPU-direct MPI. GPU-direct MPI does not need to be enabled with a specific command. It is sufficient to pass a pointer allocated with `cudaMalloc` or `hipMalloc` etc to MPI.

By the way, you still need to set the environment variable `MPICH_GPU_SUPPORT_ENABLED` to make it work, otherwise you get an error. See https://dci.dci-gitlab.cines.fr/webextranet/software_stack/libraries/index.html#gpu-aware-gpu-direct

tpadioleau:
Actually with #245 I see that we already rely on MPI GPU-aware implementation.

tpadioleau (approved):
Seems ok to me

tpadioleau (changes requested):
Just remembered the question of the modules and environment variables that should be set. If i understand the Adastra documentation we need to set `MPICH_GPU_SUPPORT_ENABLED=1` in our environment.sh ?

tpadioleau:
Given the current PR is `MPICH_GPU_SUPPORT_ENABLED` still an optimization parameter ? Would gyselalibxx run fine without setting this variable ? If not then it is not anymore an optimization parameter.

tpadioleau:
I would be in favor of putting it in the `environment.sh` file, that would simplify the use of `salloc/srun` without submission scripts.

tpadioleau:
> Do you know which machines are using mpich? I think these environment variables depend on the machine.

Cray supercomputers like Adastra, Lumi or Frontier. I think Jean-Zay uses intel mpi or openmpi.

> It would be good to merge this PR so either we need to identify these variables or merge for now and add them as we work out what they are on each machine

yes I think it is ready to be reviewed.

EmilyBourne:
> > It would be good to merge this PR so either we need to identify these variables or merge for now and add them as we work out what they are on each machine
> 
> yes I think it is ready to be reviewed.

The tests are passing but I notice you said reviewed not merged so I am waiting for your review

EmilyBourne:
The MPI tests on Adastra are also reported as failing in #340 so this does not seem to be related

tpadioleau:
> The MPI tests on Adastra are also reported as failing in #340 so this does not seem to be related

If I launch it by hand it passes:
```
[gen2224] tpadioleau@login2:~/gyselalibxx$ srun -n 2 "/lus/home/CT5/gen2224/tpadioleau/gyselalibxx/build/tests/mpi_parallelisation/unit_tests_parallelisation" "--gtest_filter=MPIParallelisation.AllToAll2D_GPU"
Note: Google Test filter = MPIParallelisation.AllToAll2D_GPU
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from MPIParallelisation
[ RUN      ] MPIParallelisation.AllToAll2D_GPU
[       OK ] MPIParallelisation.AllToAll2D_GPU (1 ms)
[----------] 1 test from MPIParallelisation (1 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1 ms total)
[  PASSED  ] 1 test.
[CRAYBLAS_WARNING] Application linked against multiple cray-libsci libraries
[CRAYBLAS_WARNING] Application linked against multiple cray-libsci libraries
```

PR #188, opened 2025-03-17
Add convergence test for partial derivatives classes

yanmnc:


tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -15,164 +17,439 @@ struct X
...
+     *
+     * @return The value of the function at the coordinate.
+     */
+    KOKKOS_FUNCTION double operator()(CoordFull const coord) const
+    {
+        double const x = ddc::get<X>(coord);
+        double const y = ddc::get<Y>(coord);
+        return (ipow(x, m_degree + 1) + ipow(y, m_degree)) * y;
+    }
+
```
EmilyBourne, line 58: We may want to consider a little more randomness. E.g. see:
https://github.com/CExA-project/ddc/blob/main/tests/splines/polynomial_evaluator.hpp

EmilyBourne: I am ok with this though

tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -15,164 +17,439 @@ struct X
...
-        = ddc::GrevilleInterpolationPoints<BSplinesY, SplineYBoundary, SplineYBoundary>;
-struct GridY : SplineInterpPointsY::interpolation_discrete_dimension_type
+
+/**
+ * @brief A class that represents a test for partial derivatives.
+ * The test can be used with finite difference method for
+ * computing partial derivatives.
+ */
+template <
+        class DerivativeDimension,
```
EmilyBourne, line 327: Is the `DerivativeDimension` necessary as a template argument? I guess it is needed by `FunctionToDifferentiatePolynomial` but not necessarily by this class if we want to be able to test different functions?

yanmnc: I think you still need to pass the `DerivativeDimension` type as a template parameter of this class given that `operator()` instantiates a member of a class that inherits from IPartialDerivativeCreator, and this latter class is templated by the dimension along which we want to evaluate partial derivatives

EmilyBourne: You could template the operator instead

tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -15,164 +17,439 @@ struct X
...
+
+public:
+    struct BSplinesX : ddc::NonUniformBSplines<X, spline_degree>
+    {
+    };
+
+    auto static constexpr SplineBoundary = ddc::BoundCond::GREVILLE;
+
+    using SplineInterpPointsX
+            = ddc::GrevilleInterpolationPoints<BSplinesX, SplineBoundary, SplineBoundary>;
```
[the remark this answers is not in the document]
EmilyBourne: Why does the FDM tests need to rely on splines though? Wouldn't it make more sense to have this in the splines and use random points for FDM?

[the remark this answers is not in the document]
yanmnc: should be fixed by 10eda0d44dafa5b50b8122394034565a2d6528e3

tests/math_tools/test_partial_derivatives.cpp
yanmnc: should be fixed now

tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -15,164 +17,447 @@ struct X
...
+    }
+
+    double const compute_error(double& max_distance) const
+    {
+        IdxRangeX const idxrange_x = SplineInterpPointsX::template get_domain<GridX>();
+        IdxRangeY const idxrange_y = SplineInterpPointsY::template get_domain<GridY>();
+        IdxRangeXY const idxrange = IdxRangeXY(idxrange_x, idxrange_y);
+
+        double dmin, dmax;
+        if (std::is_same_v<DDim, X>) {
```
EmilyBourne, line 291:
```suggestion
        if constexpr (std::is_same_v<DDim, X>) {
```

EmilyBourne:
image

EmilyBourne:
image

image

image

image

image

image

tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -15,164 +17,447 @@ struct X
...
+using CoordXY = Coord<X, Y>;
+
+
+/**
+ * @brief A class that represents a test function for computing partial derivatives.
+ * The test function is defined as the product of two cosine functions.
+ */
+class FunctionToDifferentiateCosine
 {
+public:
```
EmilyBourne, line 41:
```suggestion
public:
    KOKKOS_DEFAULTED_FUNCTION FunctionToDifferentiateCosine() = default;

```

tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -15,164 +17,463 @@ struct X
...
+    /**
+     * @brief Default copy constructor.
+     */
+    KOKKOS_DEFAULTED_FUNCTION FunctionToDifferentiateCosine(FunctionToDifferentiateCosine const&)
+            = default;
+
+    /**
+     * @brief Default destructor.
+     */
+    ~FunctionToDifferentiateCosine() = default;
```
EmilyBourne, line 56:
```suggestion
    KOKKOS_DEFAULTED_FUNCTION ~FunctionToDifferentiateCosine() = default;
```

tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -15,164 +17,463 @@ struct X
...
+ * @brief A class that represents a test function for computing partial derivatives.
+ * The test function is defined as the product of two cosine functions.
+ */
+class FunctionToDifferentiateCosine
 {
+public:
+    /**
+     * @brief Default constructor.
+     */
+    FunctionToDifferentiateCosine() = default;
```
EmilyBourne, line 45:
```suggestion
    KOKKOS_DEFAULTED_FUNCTION FunctionToDifferentiateCosine() = default;
```

yanmnc: 89e351dbb41dbe018c4b8c81e0da1a50d7954bba

EmilyBourne (approved):

PR #196, opened 2025-03-19
Replace the ".uid" in polarpoissonlikesolver.hpp.

PaulineVidal:
I noticed that the static_analysis often fails  on the polarpoissonlikesolver.hpp. because of the remaining `.uid()`. I tried to replace them.

### DIFF
--- a/src/pde_solvers/polarpoissonlikesolver.hpp
+++ b/src/pde_solvers/polarpoissonlikesolver.hpp
@@ -221,7 +221,7 @@ private:
     // Matrix size is equal to the number of Polar bspline
     const int m_matrix_size;
 
-    // Domains
+    // Index ranges
     IdxRangeBSPolar m_idxrange_fem_non_singular;
     IdxRangeBSR m_idxrange_bsplines_r;
     IdxRangeBSTheta m_idxrange_bsplines_theta;
@@ -229,6 +229,10 @@ private:
     IdxRangeQuadratureR m_idxrange_quadrature_r;
     IdxRangeQuadratureTheta m_idxrange_quadrature_theta;
     IdxRangeQuadratureRTheta m_idxrange_quadrature_singular;
+
+    // Index
+    Idx<BSplinesR> m_idx_bsplines_r_front;
+    Idx<BSplinesTheta> m_idx_bsplines_theta_front;
 
     // Gauss-Legendre points and weights
     FieldMem<double, IdxRangeQuadratureR> m_weights_r;
@@ -302,6 +306,8 @@ public:
                   m_idxrange_quadrature_r.take_first(
                           IdxStep<QDimRMesh> {m_n_overlap_cells * s_n_gauss_legendre_r}),
                   m_idxrange_quadrature_theta)
+        , m_idx_bsplines_r_front(ddc::discrete_space<BSplinesR>().full_domain().front())
+        , m_idx_bsplines_theta_front(ddc::discrete_space<BSplinesTheta>().full_domain().front())
         , m_weights_r(m_idxrange_quadrature_r)
         , m_weights_theta(m_idxrange_quadrature_theta)
         , m_singular_basis_vals_and_derivs(IdxRange<PolarBSplinesRTheta, QDimRMesh, QDimThetaMesh>(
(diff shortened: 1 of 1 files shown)


src/pde_solvers/polarpoissonlikesolver.hpp
tpadioleau, line 1: Indeed this should silence the warnings but I am afraid we still have the same underlying problem. Calling the constructor https://ddc.mdls.fr/classddc_1_1DiscreteElement.html#af6776e89752caf2077e70f690502a5e5 should be done as a last resort. One should see `uid()` and this constructor as DDC exits.

Isn't there a function in splines that could get you the `Idx<RCellDim>` for example ? Is `RCellDim` the dimension of knots ?

EmilyBourne: `RCellDim` is the dimension of the cells on which the bsplines are defined. It is declared here: 
https://github.com/gyselax/gyselalibxx/blob/0647e3d788c10a7840804ca5853b331dd44a79e2/src/pde_solvers/polarpoissonlikesolver.hpp#L80-L82

As it is inside this class I don't think it is dangerous to call this constructor.

That said in the places where ` - Idx<RCellDim>(0)` is used it seems to me that most of the operations could be carried out directly with an `Idx`. I see no reason to use integers

src/pde_solvers/polarpoissonlikesolver.hpp
```diff
@@ -1442,10 +1467,12 @@ class PolarSplineFEMPoissonLikeSolver
                     Idx<RBasisSubset> ib_trial_r(idx_trial_r - cell_idx_r);
                     Idx<ThetaBasisSubset> ib_trial_theta(theta_mod(ib_trial_theta_idx));
 
-                    assert(ib_test_r.uid() < BSplinesR::degree() + 1);
-                    assert(ib_test_theta.uid() < BSplinesTheta::degree() + 1);
-                    assert(ib_trial_r.uid() < BSplinesR::degree() + 1);
-                    assert(ib_trial_theta.uid() < BSplinesTheta::degree() + 1);
+                    assert((ib_test_r - Idx<RBasisSubset>(0)).value() < BSplinesR::degree() + 1);
```
EmilyBourne, line 1470:
```suggestion
                    assert(ib_test_r < Idx<RBasisSubset>(BSplinesR::degree() + 1));
```

src/pde_solvers/polarpoissonlikesolver.hpp
```diff
@@ -636,15 +642,20 @@ class PolarSplineFEMPoissonLikeSolver
...
-                                : idx_trial_r.uid() - BSplinesR::degree());
+                const std::size_t idx_trial_r_val = (idx_trial_r - m_idx_bsplines_r_front).value();
+                const std::size_t idx_trial_theta_val
+                        = (idx_trial_theta - m_idx_bsplines_theta_front).value();
+                const std::size_t first_overlap_element_r_val
+                        = idx_trial_r_val < BSplinesR::degree()
+                                  ? 0
+                                  : idx_trial_r_val - BSplinesR::degree();
+
+                const Idx<RCellDim> first_overlap_element_r(first_overlap_element_r_val);
```
EmilyBourne, line 653: Are integers needed here? Is it not possible to do something like:
```suggestion
const Idx<RCellDim> first_overlap_element_r(
                        idx_trial_r < Idx<RCellDim>(BSplinesR::degree())
                                ? Idx<RCellDim>(0)
                                : idx_trial_r - BSplinesR::degree());
```
?
Or is `idx_trial_r` the wrong type for this?

EmilyBourne:
If you want to fix the warnings in this file then the 2 remaining warnings (which are easier to fix) should be addressed.
That said I was under the impression that we had agreed to leave this file as it is (with the warnings) for now as this code should be replaced in the coming months by a new implementation written by @etiennemlb . I think that a clean solution to these warnings involves more work than simply writing `(ddc::select<BSplinesR>(idx_test) - m_idx_bsplines_r_front).value()` instead of `ddc::select<BSplinesR>(idx_test).uid()` to ensure that the index types are used as intended

src/pde_solvers/polarpoissonlikesolver.hpp
PaulineVidal: Indeed, `idx_trial_r` is a `Idx<BSplinesR>` and `first_overlap_element_r` a `Idx<RCellDim>`.

PaulineVidal: Done in https://github.com/gyselax/gyselalibxx/pull/196/commits/9092bfe093b512a5f0005863e77264a7a001ad16

src/pde_solvers/polarpoissonlikesolver.hpp
PaulineVidal: Done in https://github.com/gyselax/gyselalibxx/pull/196/commits/9092bfe093b512a5f0005863e77264a7a001ad16

EmilyBourne:
After discussion, it seems that this file will not be modified by Etienne in the near future so it could be helpful to clean it up. That said, the changes in this PR do not always improve the situation. There are multiple examples where a more optimal solution could be found (e.g. saving `x_init_host` in a `DField<IdxRange<BatchDim, PolarBSplines>>`).

Each of these changes should therefore be examined to find the best solution for the individual case. It may be simpler for me or @tpadioleau to do this in a separate PR if you don't mind @PaulineVidal ?

EmilyBourne:
cc @alex-m-h

EmilyBourne:
Superseded by #391 .
@PaulineVidal Do you want me to delete your branch or can you take care of it?

PaulineVidal:
You can delete it. :+1:

PR #197, opened 2025-03-21
Change input/output grid name for compatibility with netCDF

gdgirard:
Change the reading of the grid to read and write tor1 instead of grid_tor1.

### DIFF
--- a/src/io/input.hpp
+++ b/src/io/input.hpp
@@ -96,7 +96,7 @@ inline IdxRange<Grid1D> init_spline_dependent_idx_range(
                     "read_" + mesh_identifier,
                     breakpoints_name,
                     breakpoints,
-                    grid_name,
+                    mesh_identifier,
                     mesh);
             ddc::init_discrete_space<BSplines>(breakpoints);
         }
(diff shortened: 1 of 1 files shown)


src/io/input.hpp
```diff
@@ -86,17 +86,17 @@ inline IdxRange<Grid1D> init_spline_dependent_idx_range(
...
+            std::string grid_name = "grid_" + mesh_identifier;
             PDI_get_arrays("read_" + mesh_identifier, grid_name, mesh);
         } else {
             std::string breakpoints_name = "breakpoints_" + mesh_identifier;
             PDI_get_arrays(
                     "read_" + mesh_identifier,
                     breakpoints_name,
                     breakpoints,
-                    grid_name,
+                    mesh_identifier,
```
EmilyBourne, line 99: Is it normal that the keyword being read has changed but none of the `.yaml.hpp` files have been changed?

Presumably this change will also affect Voice?

gdgirard: Yes because this is a case only used for Gysela-Axi

EmilyBourne (approved):

PR #203, opened 2025-03-24
Add coverage monitoring

EmilyBourne:
Add a connection to codecov to track the code coverage in the src/ directory. This helps spot code that is added without tests and the test can be made compulsory to force complete testing.


EmilyBourne:
To complete after #248

PR #204, opened 2025-03-25
Adding of the mapping from toroidal to cylindrical coordinates.

gdgirard:


src/mapping/toroidal_to_cylindrical.hpp
```diff
@@ -0,0 +1,442 @@
...
+ *
+ * @f$ R(\rho,\theta) = \sum_k c_{R,k} B_k(\rho,\theta),@f$
+ *
+ * @f$ Z(\rho,\theta) = \sum_k c_{Z,k} B_k(\rho,\theta).@f$
+ *
+ */
+template <
+        class R,
+        class Z,
+        class Zeta class SplineEvaluator,
```
EmilyBourne, line 41:
```suggestion
        class Zeta, class SplineEvaluator,
```

gdgirard: So strange that it compiles ?

EmilyBourne: It does?! Are you sure you pushed everything?

src/mapping/toroidal_to_cylindrical.hpp
```diff
@@ -0,0 +1,442 @@
...
+    using BSplineTheta = typename SplineEvaluator::bsplines_type2;
+
+    /// @brief Indicate the first physical coordinate.
+    using cartesian_tag_R = R;
+    /// @brief Indicate the second physical coordinate.
+    using cartesian_tag_Z = Z;
+    /// @brief Indicate the first logical coordinate.
+    using curvilinear_tag_rho = Rho;
+    /// @brief Indicate the second logical coordinate.
+    using curvilinear_tag_theta = Theta;
```
EmilyBourne, line 66: We need to have a think about these names. They are sometimes used in other files. E.g here:
https://github.com/gyselax/gyselalibxx/blob/95f8db61f1440a54589d1452061b4461bc3d0312/src/interpolation/polar_splines/polar_bsplines.hpp#L218-L219

so we may need a global naming convention for this or a better way to specify this in the files where it is used

src/mapping/toroidal_to_cylindrical.hpp
```diff
@@ -0,0 +1,442 @@
...
+    using Phi_cov = typename Phi::Dual;
+
+private:
+    using spline_idx_range = IdxRange<BSplineRho, BSplineTheta>;
+
+    using SplineType = DConstField<spline_idx_range, MemorySpace>;
+
+    using IdxRangeRhoTheta = typename SplineEvaluator::evaluation_domain_type;
+    using IdxRangeTheta = typename SplineEvaluator::evaluation_domain_type2;
+    using IdxTheta = typename IdxRangeTheta::discrete_element_type;
```
EmilyBourne, line 93:
```suggestion
    using IdxRangeSpline = IdxRange<BSplineRho, BSplineTheta>;

    using SplineType = DConstField<spline_idx_range, MemorySpace>;

    using IdxRangeRhoTheta = typename SplineEvaluator::evaluation_domain_type;
    using IdxRangeTheta = typename SplineEvaluator::evaluation_domain_type2;
    using IdxTheta = typename IdxRangeTheta::discrete_element_type;
```
?

src/mapping/toroidal_to_cylindrical.hpp
```diff
@@ -0,0 +1,442 @@
...
+public:
+    /**
+     * @brief Instantiate a DiscreteToCartesian from the coefficients of 2D splines approximating the mapping.
+     *
+     * A discrete mapping is a mapping whose values are known only at the mesh points of the grid.
+     * To interpolate the mapping, we use B-splines. The DiscreteToCartesian mapping is initialised
+     * from the coefficients in front of the basis splines which arise when we approximate the
+     * functions @f$ R(\rho,\theta) @f$, and @f$ Z(\rho,\theta) @f$ (with @f$ R @f$ and @f$ Z @f$ the physical dimensions in
+     * the logical domain) with Splines (using SplineBuilder2D). Then to interpolate the mapping,
+     * we will evaluate the decomposed functions on B-splines (see DiscreteToCartesian::operator()).
```
EmilyBourne, line 110: Copy/pasted from `DiscreteToCartesian`

src/mapping/toroidal_to_cylindrical.hpp
```diff
@@ -0,0 +1,442 @@
...
+            SplineType curvilinear_to_R,
+            SplineType curvilinear_to_Z,
+            SplineEvaluator const& evaluator,
+            IdxRangeRTheta idx_range_singular_point)
+        : m_R_spline_representation(curvilinear_to_R)
+        , m_Z_spline_representation(curvilinear_to_Z)
+        , m_spline_evaluator(evaluator)
+        , m_idx_range_singular_point(idx_range_singular_point)
+    {
+    }
```
EmilyBourne, line 142: Apart from an orthogonal direction, what does this class add on top of `DiscreteToCartesian`? Is it not possible to create a:
```cpp
template <
        class DiscreteToCartesianMapping,
        class Zeta>
class ToroidalToCylindrical;
```
to avoid duplicating the logic in `DiscreteToCartesian`?

EmilyBourne: Or to extend `DiscreteToCartesian`?
```
template <
        class R,
        class Z,
        class Zeta,
        class SplineEvaluator,
        class Rho = typename SplineEvaluator::continuous_dimension_type1,
        class Theta = typename SplineEvaluator::continuous_dimension_type2,
        class MemorySpace = typename SplineEvaluator::memory_space>
class ToroidalToCylindrical : public DiscreteToCartesian<R, Z, SplineEvaluator, Rho, Theta, MemorySpace>
```

src/mapping/toroidal_to_cylindrical.hpp
```diff
@@ -0,0 +1,442 @@
...
+     * @return The el-th control point.
+     *
+     * @see GrevilleInterpolationPoints
+     * @see KnotsAsInterpolationPoints
+     */
+    KOKKOS_INLINE_FUNCTION const Coord<R, Z> control_point(
+            Idx<BSplineRho, BSplineTheta> const& el) const
+    {
+        return Coord<R, Z>(m_R_spline_representation(el), m_Z_spline_representation(el));
+    }
```
EmilyBourne, line 410: Does this method make sense for this class?

src/mapping/toroidal_to_cylindrical.hpp
```diff
@@ -0,0 +1,442 @@
...
+        DiscreteToCartesian<R, Z, SplineEvaluator, Rho, Theta, MemorySpace>>
+{
+    static constexpr bool value = Kokkos::SpaceAccessibility<ExecSpace, MemorySpace>::accessible;
+};
+
+template <class R, class Z, class SplineEvaluator, class Rho, class Theta, class MemorySpace>
+struct IsCurvilinear2DMapping<DiscreteToCartesian<R, Z, SplineEvaluator, Rho, Theta, MemorySpace>>
+    : std::true_type
+{
+};
```
EmilyBourne, line 434: This does not seem to be documented, but I think we use `IsCurvilinear2DMapping` to determine whether the following tags are defined:
```
cartesian_tag_x
cartesian_tag_y
curvilinear_tag_r
curvilinear_tag_theta
```

It certainly asks if the mapping is 2D which is not the case here

PR #193, opened 2025-03-19
MkDocs Integration into Gyselalibxx

Philipp137:
### **Summary**  
This PR introduces [**MkDocs**](https://www.mkdocs.org/) for documentation, replacing the previous Doxygen-only setup.  

### **Why MkDocs instead of Sphinx or Doxygen?**  
We chose **MkDocs** for the following reasons:  
✅ **Markdown-native**: Unlike Sphinx, which primarily uses `.rst` files, MkDocs supports Markdown `.md` natively, making integration seamless.  
✅ **Easy setup**: Uses a simple `mkdocs.yml` configuration.  
✅ **Doxygen integration**: Supports embedding API documentation via `mkdoxy`.  
✅ **Modern and customizable**: MkDocs provides a cleaner and more modern UI compared to Doxygen.  
✅ **GitHub Pages-friendly**: Simplifies deployment for project documentation.  

### **Main Changes**  
🔹 Integrated **MkDocs** into `docs/`.  
🔹 Moved previous `doc/` directory to `docs/latex/`.  
🔹 Updated **GitHub deployment**

PR #208, opened 2025-03-25
Move doc folder into docs/latex

EmilyBourne:


EmilyBourne:
This was originally part of #193 but can be done independently.

cc: @Philipp137

yanmnc (approved):

PR #215, opened 2025-03-28
Update the documentation to include the formula for the divergence operator

yanmnc:


docs/mathematical_and_physical_conventions.md
```diff
@@ -253,3 +253,25 @@ This quantity can be expressed in the both covariant $`\{\mathbf{b}^i\}`$ and co
...
+
+### Divergence
+
+The divergence of a vector field $`\mathbf{A} = A^i \mathbf{b}_i`$ is written as
+
+```math
+\nabla \cdot \mathbf{A} = \frac{1}{\sqrt{g}}\frac{\partial}{\partial q^i} \left( \sqrt{g} A^i\right), 
+```
+
+Where $`g`$ stands for the determinant of the metric tensor. Also note that $`g = \text{det}(G) = \text{det}(J^T J) = \text{det}(J)^2`$ thus $`\sqrt{g}`$ reduces to the determinant of the Jacobian tensor. In the code we evaluate the above formula by writing
```
EmilyBourne, line 265:
```suggestion
Where $`g`$ stands for the determinant of the metric tensor. Also note that $`g = \text{det}(G) = \text{det}(J^T J) = \text{det}(J)^2`$ thus $`\sqrt{g}`$ reduces to the absolute value of the determinant of the Jacobian tensor. In the code we evaluate the above formula by writing
```

PR #223, opened 2025-03-29
Improve CI automisation

EmilyBourne:
- Improve CPU/GPU test skipping by using the PR's "Changes" tab to determine relevant changes. If some are found *then* we look for duplicate runs. Otherwise these tests are always skipped. This should prevent files being detected as modified due to a merge with main.
- Fix failing clean up test which should add "Ready to review" flag by correctly identifying PR number (renamed to check_required_test_result). This workflow is actually only called after the "GPU test" so an action "set_pr_status" is also added to add the flag and/or set the draft status when jobs finish.
- Split GPU tests to avoid needing to approve a skipped test
- Remove `set_draft_cancelled` job. This causes problems more often than it fixes them
- Make GPU tests wait for Static analysis tests before requesting deployment
- Cancel running tests when the PR is put into draft
- Fix the action which removes the "Ready to review" flag on draft PRs
- Add a job to add "Ready to merge" to PRs that have been approved and to put PRs in draft if changes are requested


gdgirard (approved):
Not the good person to ensure that all is OK but I agree with the philosophy of this pull request

PR #224, opened 2025-03-29
Add a templated function to compute the components of the inverse Jacobian matrix

gdgirard:


gdgirard:
@EmilyBourne @PaulineVidal I have a problem with this pull request for the Czarny mapping. Can you check if you see something that I haven't don't correctly. 
image

Something that seems strange to me is these lines that I haven't change
```
    /**
     * @brief Get the inverse mapping.
     *
     * @return The inverse mapping.
     */
    CartesianToCzarny<X, Y, R, Theta> get_inverse_mapping() const
    {
        return CartesianToCzarny<X, Y, R, Theta>(m_epsilon, m_e);
    }
```

EmilyBourne:
> Something that seems strange to me is these lines that I haven't change
> 
> ```
>     /**
>      * @brief Get the inverse mapping.
>      *
>      * @return The inverse mapping.
>      */
>     CartesianToCzarny<X, Y, R, Theta> get_inverse_mapping() const
>     {
>         return CartesianToCzarny<X, Y, R, Theta>(m_epsilon, m_e);
>     }
> ```

This method seems fine to me, as long as it is found in the class `CzarnyToCartesian`. `CartesianToCzarny` is the inverse mapping of `CzarnyToCartesian`.

EmilyBourne:
This seems to be the case:
https://github.com/gyselax/gyselalibxx/blob/a3b82dfe16244b47d797b15ad02807b4ccfc4e00/src/mapping/czarny_to_cartesian.hpp#L441-L444

src/mapping/inverse_jacobian_matrix.hpp
```diff
@@ -73,94 +67,49 @@ class InverseJacobianMatrix
...
-     * @param[in] coord
-     * 				The coordinate where we evaluate the inverse Jacobian matrix.
-     *
-     * @return A double with the value of the (2,2) coefficient of the inverse Jacobian matrix.
-     */
-    KOKKOS_INLINE_FUNCTION double inv_jacobian_22(PositionCoordinate const& coord) const
-    {
         if constexpr (has_2d_inv_jacobian_v<Mapping, PositionCoordinate>) {
-            return m_mapping.inv_jacobian_22(coord);
+            return m_mapping.inv_jacobian_component(coord);
```
EmilyBourne, line 88:
```suggestion
            return m_mapping.template inv_jacobian_component<IndexTag1, IndexTag2>(coord);
```

tests/mapping/mapping_jacobian_matrix_coef.cpp
```diff
@@ -193,19 +199,24 @@ TEST_P(JacobianMatrixAndJacobianCoefficients, MatrixDiscCzarMap)
...
+            std::cout << "1: " << ddcHelper::get<R, X_cov>(inv_Jacobian_matrix_coeff) << " : " << ddcHelper::get<R,X_cov>(inv_Jacobian_matrix) << std::endl;
+            ddcHelper::get<R, Y_cov>(inv_Jacobian_matrix_coeff)
+                    = inv_jacobian.template inv_jacobian_component<R, Y_cov>(coord_rtheta);
+            std::cout << "2: " << ddcHelper::get<R, Y_cov>(inv_Jacobian_matrix_coeff) << " : " << ddcHelper::get<R,Y_cov>(inv_Jacobian_matrix) << std::endl;
+            ddcHelper::get<Theta, X_cov>(inv_Jacobian_matrix_coeff)
+                    = inv_jacobian.template inv_jacobian_component<Theta, X_cov>(coord_rtheta);
+            std::cout << "3: " << ddcHelper::get<Theta, X>(inv_Jacobian_matrix_coeff) << " : " << ddcHelper::get<Theta,X_cov>(inv_Jacobian_matrix) << std::endl;
+            ddcHelper::get<Theta, Y_cov>(inv_Jacobian_matrix_coeff)
+                    = inv_jacobian.template inv_jacobian_component<Theta, Y_cov>(coord_rtheta);
+            std::cout << "4: " << ddcHelper::get<Theta, Y_cov>(inv_Jacobian_matrix_coeff) << " : " << ddcHelper::get<Theta,Y_cov>(inv_Jacobian_matrix) << std::endl;
```
EmilyBourne, line 219:
```suggestion
            ddcHelper::get<R, Y_cov>(inv_Jacobian_matrix_coeff)
                    = inv_jacobian.template inv_jacobian_component<R, Y_cov>(coord_rtheta);
            ddcHelper::get<Theta, X_cov>(inv_Jacobian_matrix_coeff)
                    = inv_jacobian.template inv_jacobian_component<Theta, X_cov>(coord_rtheta);
            ddcHelper::get<Theta, Y_cov>(inv_Jacobian_matrix_coeff)
                    = inv_jacobian.template inv_jacobian_component<Theta, Y_cov>(coord_rtheta);
```

EmilyBourne, line 219: Given the output that you have here, it looks like you are always ending up in the `else` condition

src/mapping/inverse_jacobian_matrix.hpp
```diff
@@ -73,94 +67,49 @@ class InverseJacobianMatrix
...
+                    std::is_same_v<IndexTag1, DimRes1> && std::is_same_v<IndexTag2, DimArg0_cov>) {
+                //Compute the (2,1) coefficient of the inverse Jacobian matrix.
+                // J^{-1}(2,1) = -J(2,1) / det(J)
+                return -m_mapping.template jacobian_component<DimRes1, DimArg0_cov>(coord) / jacob;
+            } else {
+                //Compute the (2,1) coefficient of the inverse Jacobian matrix.
+                // J^{-1}(2,2) = J(1,1) / det(J)
+                return m_mapping.template jacobian_component<DimRes0, DimArg0_cov>(coord) / jacob;
+            }
         }
```
EmilyBourne, line 113: The tags for the inverse and the tags for the jacobian should be complementary not identical. See the static assert above for the tags.
```suggestion
            static_assert(
            if constexpr (
                    std::is_same_v<IndexTag1, DimArg0> && std::is_same_v<IndexTag2, DimRes0_cov>) {
                //Compute the (1,1) coefficient of the inverse Jacobian matrix.
                // J^{-1}(1,1) = J(2,2) / det(J)
                return m_mapping.template jacobian_component<DimRes1, DimArg1_cov>(coord) / jacob;
            } else if constexpr (
                    std::is_same_v<IndexTag1, DimArg0> && std::is_same_v<IndexTag2, DimRes1_cov>) {
                //Compute the (1,2) coefficient of the inverse Jacobian matrix.
                // J^{-1}(1,2) = -J(1,2) / det(J)
                return -m_mapping.template jacobian_component<DimRes0, DimArg1_cov>(coord) / jacob;
            } else if constexpr (
                    std::is_same_v<IndexTag1, DimArg1> && std::is_same_v<IndexTag2, DimRes0_cov>) {
                //Compute the (2,1) coefficient of the inverse Jacobian matrix.
                // J^{-1}(2,1) = -J(2,1) / det(J)
                return -m_mapping.template jacobian_component<DimRes1, DimArg0_cov>(coord) / jacob;
            } else {
                //Compute the (2,1) coefficient of the inverse Jacobian matrix.
                // J^{-1}(2,2) = J(1,1) / det(J)
                return m_mapping.template jacobian_component<DimRes0, DimArg0_cov>(coord) / jacob;
            }
        }
```

EmilyBourne:
> I have a problem with this pull request for the Czarny mapping

The problem is related to a discrete mapping not a Czarny mapping

src/mapping/inverse_jacobian_matrix.hpp
gdgirard: How can my line 88 compiles ?

EmilyBourne: I am not sure. This seems to be compiler dependant. It is also possible that it doesn't compile but is not tested, the `if constexpr` would mean that it was never compiled in this case. This can be confirmed when I have finished #203

tests/mapping/mapping_jacobian_matrix_coef.cpp
gdgirard: Yes it is what I had in mind but I believed that the problem was in Czarny mapping.

src/mapping/circular_to_cartesian.hpp
```diff
@@ -228,74 +221,45 @@ class CircularToCartesian
...
+            //Compute the (2,1) coefficient of the inverse Jacobian matrix.
+            const double r = ddc::get<R>(coord);
+            assert(fabs(r) >= 1e-15);
+            return -1 / r * Kokkos::sin(theta);
+        } else {
+            //Compute the (2,2) coefficient of the inverse Jacobian matrix.
+            const double r = ddc::get<R>(coord);
+            assert(fabs(r) >= 1e-15);
+            return 1 / r * Kokkos::cos(theta);
+        }
```
EmilyBourne, line 259: Code duplication could be reduced. I'm not sure that this reduces the number of lines though so feel free to pick the version that you find more readable
```suggestion
        } else {
            const double r = ddc::get<R>(coord);
            assert(fabs(r) >= 1e-15);
            if constexpr (std::is_same_v<IndexTag2, X_cov>) {
                //Compute the (2,1) coefficient of the inverse Jacobian matrix.
                return -1 / r * Kokkos::sin(theta);
            } else {
                //Compute the (2,2) coefficient of the inverse Jacobian matrix.
                return 1 / r * Kokkos::cos(theta);
            }
        }
```

src/mapping/combined_mapping.hpp
```diff
@@ -221,52 +221,23 @@ class CombinedMapping
...
+    template <class IndexTag1, class IndexTag2>
+    KOKKOS_INLINE_FUNCTION double inv_jacobian_component(CoordJacobian const& coord_rtheta) const
     {
-        InvJacobianMatrixType J = inv_jacobian_matrix(coord_rtheta);
-        return ddcHelper::get<DimArg1, typename DimResult2::Dual>(J);
-    }
+        static_assert(ddc::in_tags_v<IndexTag1, VectorIndexSet<DimArg1, DimArg2>>);
+        static_assert(ddc::in_tags_v<
+                      IndexTag2,
+                      VectorIndexSet<typename DimResult1::Dual, typename DimResult2::Dual>>);
```
EmilyBourne, line 235: I don't know if these static assertions are very readable?
In any case the line `ddcHelper::get<IndexTag1, IndexTag2>(J);` will raise a static_assert if the tags are not present in the inverse Jacobian matrix.
If you think they are easy to understand then lets keep them. If not then maybe the assertions in the getter are sufficient. Removing them would also mean that only 1 function (`inv_jacobian_matrix`) is specific to the 2d case. This may be helpful as we move towards higher dimensional mappings

tests/mapping/mapping_jacobian_matrix_coef.cpp
```diff
@@ -108,16 +110,20 @@ TEST_P(JacobianMatrixAndJacobianCoefficients, MatrixCzarMap)
 
     InverseJacobianMatrix inv_jacobian(mapping);
     // --- for the inverseJacobian matrix:
+    using X_cov = X::Dual;
+    using Y_cov = Y::Dual;
```
EmilyBourne, line 114: Usually `X_cov == X` these lines are probably unnecessary

EmilyBourne (changes requested):

tests/mapping/mapping_jacobian_matrix_coef.cpp
gdgirard: Strange for me also. But if I don't add these lines I have the following error
```
/home/vg160296/Codes/gyselalibxx_github/tests/mapping/mapping_jacobian_matrix_coef.cpp:118:58: error: ‘X_cov’ was not declared in this scope; did you mean ‘R_cov’?
  118 |         DTensor<VectorIndexSet<R, Theta>, VectorIndexSet<X_cov, Y_cov>> inv_Jacobian_matrix_coeff;
```

EmilyBourne: Yes, as you are using `X_cov` but not defining it. However as `X` and `X_cov` represent the same type you should be able to do:
```
DTensor<VectorIndexSet<R, Theta>, VectorIndexSet<X, Y>> inv_Jacobian_matrix_coeff;
```

gdgirard: Yes but I believe that it is clearer to keep the fact that the covariant variables are needed

EmilyBourne: Ok. Then I am ok with the code as is :+1:

src/mapping/combined_mapping.hpp
gdgirard: I may have added this type of line elsewhere, where it wasn't necessary ?

EmilyBourne: Yes, I think there is one in this file related to the jacobian

EmilyBourne: https://github.com/gyselax/gyselalibxx/blob/22a550e9c764bbaeee6334a42c295bc6042e9782/src/mapping/combined_mapping.hpp#L165-L169

gdgirard: Corrected with commit b2ccdfaf500.

EmilyBourne (approved):

PR #228, opened 2025-03-31
Add a Levi-Civita tensor type

### LINKED ISSUE #221: Add a Levi Civita class
Add a class to represent the Levi-Civita symbol. This class should have a constexpr `get` operation and should be interoperable with Tensor/IndexedTensor.
It should take 1 VectorIndexSet as a template parameter

EmilyBourne:
Fixes #221

### DIFF
--- a/src/data_types/static_tensors.hpp
+++ b/src/data_types/static_tensors.hpp
@@ -6,6 +6,12 @@
 #include "type_seq_tools.hpp"
 #include "vector_index_tools.hpp"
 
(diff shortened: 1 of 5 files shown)


EmilyBourne:
Godbolt shows that from -O1 using a constexpr function returning 0 in a multiplication is optimised out:
https://godbolt.org/z/6qar7fE5T

gdgirard (approved):

PR #233, opened 2025-04-01
Store advection field in Vector not Coordinate

EmilyBourne:
This PR replaces the use of `CoordXY` with `DVector<X,Y>` to store the advection field in the Cartesian coordinate system. The advection field is not a coordinate so this type is not logical (it was used previously as Vector did not yet exist). Related to #39.

### DIFF
--- a/src/geometryRTheta/advection_field/advection_field_rtheta.hpp
+++ b/src/geometryRTheta/advection_field/advection_field_rtheta.hpp
@@ -351,7 +351,7 @@ public:
      * @param[out] advection_field_rtheta
      *      The advection field on the logical axis. It is expressed on the contravariant basis. 
      * @param[out] advection_field_xy_centre
-     *      The advection field on the physical axis at the O-point. 
+     *      The advection field at the centre point on the Cartesian basis.
      */
     void operator()(
             host_t<DFieldRTheta> electrostatic_potential,
@@ -383,6 +383,8 @@ public:
      *      The advection field on the logical axis. It is expressed on the contravariant basis. 
      * @param[out] advection_field_xy_centre
      *      The advection field on the physical axis at the O-point.  
+     * @param[out] advection_field_xy_centre
+     *      The advection field at the centre point on the Cartesian basis.
      */
     void operator()(
             host_t<Spline2D> electrostatic_potential_coef,
@@ -406,7 +408,7 @@ public:
      * @param[out] advection_field_rtheta
      *      The advection field on the logical axis. It is expressed on the contravariant basis. 
      * @param[out] advection_field_xy_centre
-     *      The advection field on the physical axis at the O-point. 
+     *      The advection field at the centre point on the Cartesian basis.
      */
     void operator()(
             host_t<PolarSplineMemRTheta>& electrostatic_potential_coef,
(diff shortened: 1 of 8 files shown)


src/geometryRTheta/advection_field/advection_field_rtheta.hpp
```diff
@@ -536,6 +536,6 @@ class AdvectionFieldFinder
         double const deriv_y_phi_0
                 = (-dr_x_2 * deriv_r_phi_1 + dr_x_1 * deriv_r_phi_2) / determinant;
 
-        advection_field_xy_centre = CoordXY(-deriv_y_phi_0, deriv_x_phi_0);
+        advection_field_xy_centre = DVector<X, Y>(-deriv_y_phi_0, deriv_x_phi_0);
```
PaulineVidal, line 539: So the `DVector<X,Y>` acts as a `CoordXY`? It stores 2 doubles? And it is different as a `DVectorField`?

tests/geometryRTheta/advection_field_rtheta/advection_field_gtest.cpp
```diff
@@ -214,9 +214,10 @@ TEST(AdvectionFieldRThetaComputation, TestAdvectionFieldFinder)
         allfdistribu_xy(irtheta) = allfdistribu_rtheta(irtheta);
         electrostatic_potential(irtheta) = simulation.electrostatical_potential(coord_xy, 0);
 
-        CoordXY const evaluated_advection_field = simulation.advection_field(coord_xy, 0);
-        ddcHelper::get<X>(advection_field_exact)(irtheta) = CoordX(evaluated_advection_field);
-        ddcHelper::get<Y>(advection_field_exact)(irtheta) = CoordY(evaluated_advection_field);
+        ddcHelper::assign_vector_field_element(
```
PaulineVidal, line 217: Do you think it is possible to overload `operator+`, `operator-` ?
To have `advection_field_exact(irtheta) = simulation.advection_field(coord_xy, 0)` ?

src/geometryRTheta/advection_field/advection_field_rtheta.hpp
EmilyBourne: > So the DVector<X,Y> acts as a CoordXY? 

No it is slightly more complex than a `CoordXY`. It has more operators and can be used with the `tensor_mul` method. Internally they both contain `std::array<double, 2>` though.

> And it is different as a DVectorField?

Yes a DVectorField is a Field of Vectors, meaning that at each grid point it describes a DVector

EmilyBourne: `Vector` is an alias for a 1D tensor. You can find the docs about tensors here:
https://gyselax.github.io/gyselalibxx/gyselalibxx/classTensor.html

PaulineVidal:
Don't forget to update the documentation. I cannot comment but I have seen at least this line:

https://github.com/gyselax/gyselalibxx/pull/233/files#diff-390307dd19116aafcc188ad596b51c7dc6b306a0750b1a5291e71acca65b530aL162

EmilyBourne:
> Don't forget to update the documentation. I cannot comment but I have seen at least this line:
> 
> https://github.com/gyselax/gyselalibxx/pull/233/files#diff-390307dd19116aafcc188ad596b51c7dc6b306a0750b1a5291e71acca65b530aL162

Arghh, I will have a look and find them. Generally types don't appear in docstrings as they are already available from the prototype and aren't always very explicit so I didn't think to check this

EmilyBourne:
> Don't forget to update the documentation. I cannot comment but I have seen at least this line:
> 
> https://github.com/gyselax/gyselalibxx/pull/233/files#diff-390307dd19116aafcc188ad596b51c7dc6b306a0750b1a5291e71acca65b530aL162

Done : [523fdeb](https://github.com/gyselax/gyselalibxx/pull/233/commits/523fdeb62ed02864a13fbee96c1887b9488a8b63)

PaulineVidal (approved):

PR #236, opened 2025-04-01
Adapt `EdgeTransformation` operator to onion geometry. 

### LINKED ISSUE #62: Adapt EdgeTransformation operator to periodic case
The problem with the periodic case is that the `ddc::coordiante(idx_range.back())` do not return the maximum value on the domain.

Eg. on \\(\[0, 2\\pi\]\\) , it returns \\(2\\pi - d\\theta\\).

So the way the operator is written is not adapted to this case.

PaulineVidal:
It suppose to fix #62 and make the `EdgeTransformation` operator to work with the "onion" geometry.

### DIFF
--- a/src/multipatch/connectivity/edge_transformation.hpp
+++ b/src/multipatch/connectivity/edge_transformation.hpp
@@ -341,6 +341,8 @@ public:
                         target_idx = target_idx_range.front();
                     }
                 }
+                assert(target_idx <= target_idx_range.back());
+                assert(target_idx >= target_idx_range.front());
             }
         } else { // Non uniform case
             // Dichotomy method comparing the coordinates of indexes of the target edge.
@@ -411,6 +413,8 @@ private:
                 target_idx_max = target_idx_range.front();
             }
         }
+        assert(target_idx_max <= target_idx_range.back());
+        assert(target_idx_max >= target_idx_range.front());
 
         CurrentCoord target_coord_min
                 = transform_edge_coord<TargetPatch>(ddc::coordinate(target_idx_min));
--- a/src/multipatch/connectivity/edge_transformation.hpp
+++ b/src/multipatch/connectivity/edge_transformation.hpp
@@ -356,7 +356,7 @@ public:
      * 
      */
     template <class CurrentGrid, class TargetGrid>
-    bool search_for_match(Idx<TargetGrid>& target_idx, Idx<CurrentGrid> const& current_idx) const
+    bool search_for_match(Idx<TargetGrid>& target_idx, Idx<CurrentGrid> current_idx) const
     {
         static_assert(
                 std::is_same_v<CurrentGrid, EdgeGrid1> || std::is_same_v<CurrentGrid, EdgeGrid2>,
@@ -377,6 +377,8 @@ public:
         using IdxRangeTarget = IdxRange<TargetGrid>;
         // Coordinate on the discontinuous dimension of the current grid.
         using CurrentCoord = Coord<typename CurrentGrid::continuous_dimension_type>;
+        // Index on the current grid
+        using IdxCurrent = typename IdxRangeCurrent::discrete_element_type;
         // Index on the target grid
         using IdxTarget = typename IdxRangeTarget::discrete_element_type;
         // Index step on the target grid
@@ -395,6 +397,13 @@ public:
                 = current_idx_range.size() - int(!CurrentGrid::continuous_dimension_type::PERIODIC);
         int n_cells_target
                 = target_idx_range.size() - int(!TargetGrid::continuous_dimension_type::PERIODIC);
+
+        if constexpr (CurrentGrid::continuous_dimension_type::PERIODIC) {
+            current_idx 
+                    = (current_idx - current_idx_range.front()).value() < n_cells_current
+                              ? current_idx
+                              : IdxCurrent((current_idx - current_idx_range.back()).value() - 1);
+        }
 
         if constexpr ( // Uniform case
                 ddc::is_uniform_point_sampling_v<
@@ -422,12 +431,11 @@ public:
                     target_idx = IdxTarget(target_idx_range.front())
                                  + int(TargetGrid::continuous_dimension_type::PERIODIC)
                                  + target_idx_step;
-
-                    if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
-                        // Apply periodicity property of the domain.
-                        if (target_idx == target_idx_range.back() + 1) {
-                            target_idx = target_idx_range.front();
-                        }
+                }
+                if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
+                    // Apply periodicity property of the domain.
+                    if (target_idx == target_idx_range.back() + 1) {
+                        target_idx = target_idx_range.front();
                     }
                 }
             }
@@ -470,7 +478,7 @@ private:
         using IdxStepTarget = IdxStep<TargetGrid>;
 
         // Dichotomy method
-        CurrentCoord const current_coord(ddc::coordinate(current_idx));
+        CurrentCoord current_coord(ddc::coordinate(current_idx));
 
         IdxTarget target_idx_min(target_idx_range.front());
         IdxTarget target_idx_max(
@@ -512,6 +520,8 @@ private:
         CurrentCoord target_coord_max
                 = (*this).template operator()<TargetPatch>(ddc::coordinate(target_idx_max));
 
+        std::cout << target_coord_min << "   " << current_coord << "   " << target_coord_max
+                  << std::endl;
         if (abs(current_coord - target_coord_min) < abs(current_coord - target_coord_max)) {
             return target_idx_min;
         } else {
--- a/src/multipatch/connectivity/edge_transformation.hpp
+++ b/src/multipatch/connectivity/edge_transformation.hpp
@@ -410,10 +410,9 @@ public:
 
         // Periodicity property.
         if constexpr (CurrentGrid::continuous_dimension_type::PERIODIC) {
-            current_idx
-                    = (current_idx - current_idx_range.front()).value() < n_cells_current
-                              ? current_idx
-                              : IdxCurrent((current_idx - current_idx_range.back()).value());
+            current_idx = (current_idx - current_idx_range.front()).value() < n_cells_current
+                                  ? current_idx
+                                  : IdxCurrent((current_idx - current_idx_range.back()).value());
         }
 
         if constexpr ( // Uniform case
@@ -439,8 +438,7 @@ public:
 
                 if constexpr (!Interface::orientations_agree) {
                     IdxStepTarget target_idx_step = target_idx_range.back() - target_idx;
-                    target_idx = IdxTarget(target_idx_range.front())
-                                 + target_idx_step;
+                    target_idx = IdxTarget(target_idx_range.front()) + target_idx_step;
                 }
 
                 // Apply periodicity property of the domain.
@@ -500,7 +498,7 @@ private:
         while (target_idx_step_diff != IdxStepTarget(1)) {
             CurrentCoord target_equivalent_coord_mid
                     = (*this).template operator()<TargetPatch>(ddc::coordinate(target_idx_mid));
-            // Periodicity property. 
+            // Periodicity property.
             if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
                 if (target_idx_mid == target_idx_range.back()) {
                     target_equivalent_coord_mid = (*this).template operator()<TargetPatch>(
@@ -520,7 +518,7 @@ private:
             target_idx_step_diff = target_idx_max - target_idx_min;
         }
 
-        // Periodicity property. 
+        // Periodicity property.
         if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
             if (target_idx_max == target_idx_range.back()) {
                 target_idx_max = target_idx_range.front();
--- a/src/multipatch/connectivity/edge_transformation.hpp
+++ b/src/multipatch/connectivity/edge_transformation.hpp
@@ -408,7 +408,7 @@ private:
         using IdxStepTarget = IdxStep<TargetGrid>;
 
         // Dichotomy method
-        CurrentCoord current_coord(ddc::coordinate(current_idx));
+        CurrentCoord const current_coord(ddc::coordinate(current_idx));
 
         IdxTarget target_idx_min(target_idx_range.front());
         IdxTarget target_idx_max(target_idx_range.back());
@@ -418,6 +418,7 @@ private:
         while (target_idx_step_diff != IdxStepTarget(1)) {
             CurrentCoord target_equivalent_coord_mid
                     = (*this).template operator()<TargetPatch>(ddc::coordinate(target_idx_mid));
+            
             // Periodicity property.
             if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
                 if (target_idx_mid == target_idx_range.back()) {
--- a/src/multipatch/connectivity/edge_transformation.hpp
+++ b/src/multipatch/connectivity/edge_transformation.hpp
@@ -134,12 +134,11 @@ get_equivalent_target_coordinate(
     IdxRangeTarget const target_idx_range(combined_idx_range);
 
     Coord<CurrentDim> const current_min = ddc::coordinate(current_idx_range.front());
-    double current_length = ddcHelper::total_interval_length(current_idx_range);
+    double const current_length = ddcHelper::total_interval_length(current_idx_range);
 
     Coord<TargetDim> const target_min = ddc::coordinate(target_idx_range.front());
     double target_length = ddcHelper::total_interval_length(target_idx_range);
 
-    std::cout << current_length << "   " << target_length << std::endl;
 
     double rescale_x = (current_coord - current_min) / current_length * target_length;
 
@@ -209,12 +208,6 @@ public:
     EdgeTransformation(
             IdxRangeEdge1 const& idx_range_patch_1,
             IdxRangeEdge2 const& idx_range_patch_2)
-        //     : m_idx_range_patch_1(
-        //         idx_range_patch_1.front(),
-        //         idx_range_patch_1.extents() + int(EdgeDim1::PERIODIC))
-        // , m_idx_range_patch_2(
-        //           idx_range_patch_2.front(),
-        //           idx_range_patch_2.extents() + int(EdgeDim2::PERIODIC))
         : m_idx_range_patch_1(idx_range_patch_1)
         , m_idx_range_patch_2(idx_range_patch_2)
     {
@@ -412,11 +405,6 @@ public:
         IdxRangeCurrent const current_idx_range(combined_idx_range);
         IdxRangeTarget const target_idx_range(combined_idx_range);
 
-        // int n_cells_current
-        //         = current_idx_range.size() - int(!CurrentGrid::continuous_dimension_type::PERIODIC);
-        // int n_cells_target
-        //         = target_idx_range.size() - int(!TargetGrid::continuous_dimension_type::PERIODIC);
-
         int n_cells_current = current_idx_range.size() - 1;
         int n_cells_target = target_idx_range.size() - 1;
 
@@ -425,12 +413,8 @@ public:
             current_idx
                     = (current_idx - current_idx_range.front()).value() < n_cells_current
                               ? current_idx
-                              // : IdxCurrent((current_idx - current_idx_range.back()).value() - 1);
                               : IdxCurrent((current_idx - current_idx_range.back()).value());
         }
-
-        std::cout << n_cells_current << "   " << current_idx_range.back() << "     " << current_idx
-                  << std::endl;
 
         if constexpr ( // Uniform case
                 ddc::is_uniform_point_sampling_v<
@@ -456,12 +440,12 @@ public:
                 if constexpr (!Interface::orientations_agree) {
                     IdxStepTarget target_idx_step = target_idx_range.back() - target_idx;
                     target_idx = IdxTarget(target_idx_range.front())
-                                 // + int(TargetGrid::continuous_dimension_type::PERIODIC)
                                  + target_idx_step;
                 }
+
+                // Apply periodicity property of the domain.
                 if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
-                    // Apply periodicity property of the domain.
-                    if (target_idx == target_idx_range.back()) { // + 1) {
+                    if (target_idx == target_idx_range.back()) {
                         target_idx = target_idx_range.front();
                     }
                 }
@@ -510,17 +494,14 @@ private:
 
         IdxTarget target_idx_min(target_idx_range.front());
         IdxTarget target_idx_max(target_idx_range.back());
-        // + int(TargetGrid::continuous_dimension_type::PERIODIC));
         IdxTarget target_idx_mid = get_mid(target_idx_min, target_idx_max);
-
-        std::cout << target_idx_min << "  " << target_idx_max << std::endl;
 
         IdxStepTarget target_idx_step_diff = target_idx_max - target_idx_min;
         while (target_idx_step_diff != IdxStepTarget(1)) {
             CurrentCoord target_equivalent_coord_mid
                     = (*this).template operator()<TargetPatch>(ddc::coordinate(target_idx_mid));
+            // Periodicity property. 
             if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
-                // if (target_idx_mid == IdxTarget(target_idx_range.back()+ 1)) {
                 if (target_idx_mid == target_idx_range.back()) {
                     target_equivalent_coord_mid = (*this).template operator()<TargetPatch>(
                             ddc::coordinate(target_idx_range.front()));
@@ -539,8 +520,8 @@ private:
             target_idx_step_diff = target_idx_max - target_idx_min;
         }
 
+        // Periodicity property. 
         if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
-            // if (target_idx_max == IdxTarget(target_idx_range.back() + 1)) {
             if (target_idx_max == target_idx_range.back()) {
                 target_idx_max = target_idx_range.front();
             }
@@ -550,9 +531,6 @@ private:
                 = (*this).template operator()<TargetPatch>(ddc::coordinate(target_idx_min));
         CurrentCoord target_coord_max
                 = (*this).template operator()<TargetPatch>(ddc::coordinate(target_idx_max));
-
-        std::cout << target_idx_min << "  " << target_idx_max << "  " << target_coord_min << "  "
-                  << current_coord << "  " << target_coord_max << std::endl;
 
         if (abs(current_coord - target_coord_min) < abs(current_coord - target_coord_max)) {
             return target_idx_min;
--- a/src/multipatch/connectivity/edge_transformation.hpp
+++ b/src/multipatch/connectivity/edge_transformation.hpp
@@ -62,9 +62,6 @@ get_equivalent_target_coordinate(
 
     Coord<TargetDim> const target_min = ddc::coordinate(target_idx_range.front());
     double target_length = ddcHelper::total_interval_length(target_idx_range);
-    std::cout << "current length = " << current_length << " | target length =  " << target_length
-              << " | current min =  " << current_min << " | target min =  " << target_min
-              << std::endl;
 
     double rescale_x = (current_coord - current_min) / current_length * target_length;
 
@@ -72,10 +69,9 @@ get_equivalent_target_coordinate(
     if constexpr (!orientations_agree) {
         rescale_x = target_length - rescale_x;
     }
-    using TargetGrid
-    = std::conditional_t<std::is_same_v<CurrentDim, EdgeDim1>, EdgeGrid2, EdgeGrid1>;
-    if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
-        rescale_x = std::fmod(rescale_x, target_length);  
+    using TargetDim = std::conditional_t<std::is_same_v<CurrentDim, EdgeDim1>, EdgeDim2, EdgeDim1>;
+    if constexpr (TargetDim::PERIODIC) {
+        rescale_x = std::fmod(rescale_x, target_length);
     }
     return target_min + rescale_x;
 }
@@ -102,7 +98,6 @@ get_equivalent_target_coordinate(
         IdxRange<typename Interface::Edge2::parallel_grid> const& idx_range_patch_2)
 {
     using Patch1 = typename Interface::Edge1::associated_patch;
-    // using Patch2 = typename Interface::Edge2::associated_patch;
 
     using EdgeGrid1 = typename Interface::Edge1::parallel_grid;
     using EdgeGrid2 = typename Interface::Edge2::parallel_grid;
@@ -131,9 +126,6 @@ get_equivalent_target_coordinate(
 
     Coord<EdgeDim> const target_min = ddc::coordinate(target_idx_range.front());
     double target_length = ddcHelper::total_interval_length(target_idx_range);
-    std::cout << "current length = " << current_length << " | target length =  " << target_length
-              << " | current min =  " << current_min << " | target min =  " << target_min
-              << std::endl;
 
     double rescale_x = (current_coord - current_min) / current_length * target_length;
 
@@ -141,10 +133,8 @@ get_equivalent_target_coordinate(
     if constexpr (!orientations_agree) {
         rescale_x = target_length - rescale_x;
     }
-    using TargetGrid
-    = std::conditional_t<std::is_same_v<CurrentPatch, Patch1>, EdgeGrid2, EdgeGrid1>;
-    if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
-        rescale_x = std::fmod(rescale_x, target_length);  
+    if constexpr (EdgeDim::PERIODIC) {
+        rescale_x = std::fmod(rescale_x, target_length);
     }
     return target_min + rescale_x;
 }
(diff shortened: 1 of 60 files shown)


EmilyBourne:
@PaulineVidal you requested my review but this is still in draft. Is there a section you are working on that you want me to check in particular?

PaulineVidal:
No, just a bad habit. I always think that adding reviewers removes the "draft" status. I am not careful.
So I am suggesting this change to make it work with patches having the continuous dimension and with edge with periodic domain. I am not sure it is the cleanest way to do it. I tried to not have too many duplication and make it work.

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -12,6 +12,145 @@
 #include "interface.hpp"
 
 
+/**
+ * @brief Struct to identify the type of an index range on the current 
+ * patch and the index range type on the target patch. 
+ */
+template <class Interface, typename T>
```
EmilyBourne, line 19:
```suggestion
template <class Interface, class Patch>
```
?

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -269,16 +423,27 @@ class EdgeTransformation
                 target_idx += target_idx_step_rescaled;
 
                 if constexpr (!Interface::orientations_agree) {
-                    IdxStepTarget target_idx_step = IdxTarget(target_idx_range.back()) - target_idx;
-                    target_idx = IdxTarget(target_idx_range.front());
-                    target_idx += target_idx_step;
+                    if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
+                        IdxStepTarget target_idx_step
+                                = IdxTarget(target_idx_range.back() + 1) - target_idx;
```
EmilyBourne, line 428: I am a little lost in these calculations. Is it not simpler to calculate the `target_idx_step_rescaled` and then calculate `target_idx` in the `if constexpr`:
```cpp
if constexpr (!Interface::orientations_agree) {
    target_idx = target_idx_range.back() + int(TargetGrid::continuous_dimension_type::PERIODIC) - target_idx_step_rescaled;
} else {
    target_idx = target_idx_range.front() + target_idx_step_rescaled;
}
```
?

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -269,16 +423,27 @@ class EdgeTransformation
...
+                    }
                 }
             }
         } else { // Non uniform case
             // Dichotomy method comparing the coordinates of indexes of the target edge.
             target_idx = get_target_idx(target_idx_range, current_idx);
             CurrentCoord const current_coord(ddc::coordinate(current_idx));
-            CurrentCoord const target_equivalent_coord((*this)(ddc::coordinate(target_idx)));
+            CurrentCoord target_equivalent_coord
+                    = (*this).template operator()<TargetPatch>(ddc::coordinate(target_idx));
```
EmilyBourne, line 446: I don't like this syntax. If we expect to need to pass template parameters to a method, I usually recommend not calling it `operator()`

Why is the `TargetPatch` necessary now when it wasn't before?

EmilyBourne (changes requested):
I have done half a review, but generally this seems like a very complicated solution to a very simple problem. Could the issue not be fixed in the constructor of `EdgeTransformation` by simply adapting the idx_range that is saved to include the periodic point?

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: In the onion geometry, we gave the same continuous dimensions to all the patch. The operator cannot determine which patch is the current one and which one in the target if it only has the continuous dimension. That is why I added the Patch as template parameter. It is supposed to do the same thing as the previous one (that I kept and which can works if the continuous dimensions are different). Indicating the Patch as template parameter works is the continuous dimensions are the same or not. 
If the continuous dimensions are different, I can use the other operator() which can deduce the template parameter.

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -243,8 +391,14 @@ class EdgeTransformation
...
-        int const n_cells_current = current_idx_range.size() - 1;
-        int const n_cells_target = target_idx_range.size() - 1;
+        int n_cells_current = current_idx_range.size() - 1;
+        int n_cells_target = target_idx_range.size() - 1;
+        if constexpr (CurrentGrid::continuous_dimension_type::PERIODIC) {
+            n_cells_current++;
+        }
+        if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
+            n_cells_target++;
+        }
```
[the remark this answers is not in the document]
PaulineVidal: I thought that using `constexpr` was better.

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -269,16 +423,27 @@ class EdgeTransformation
                 target_idx += target_idx_step_rescaled;
 
                 if constexpr (!Interface::orientations_agree) {
-                    IdxStepTarget target_idx_step = IdxTarget(target_idx_range.back()) - target_idx;
-                    target_idx = IdxTarget(target_idx_range.front());
-                    target_idx += target_idx_step;
+                    if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
```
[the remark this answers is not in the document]
PaulineVidal: The problem with the PERIODIC conditions is that the last interpolation points is _not_ the boundary of the domain. So, everytime we are using `idx_range.back()` thinking that its `ddc::coordinate()` is `theta_max`, it is actually `theta_max - dtheta`. And `idx_range.back() +1` is `theta_max`.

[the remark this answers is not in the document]
EmilyBourne: Yes. But why is this only true/only a problem if the orientations don't agree?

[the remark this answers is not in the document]
PaulineVidal: If they agree, we dont call `.back()` and the number of cells were already fixed in https://github.com/gyselax/gyselalibxx/pull/236/files#diff-6ca366f244ec8fff8736b57bfc1bb1f221c1388e348bd154e2b7b484073586d0R397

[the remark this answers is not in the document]
EmilyBourne: Don't you need this though:
https://github.com/gyselax/gyselalibxx/pull/236/files#diff-6ca366f244ec8fff8736b57bfc1bb1f221c1388e348bd154e2b7b484073586d0R426-R431
?

PaulineVidal:
> I have done half a review, but generally this seems like a very complicated solution to a very simple problem. Could the issue not be fixed in the constructor of EdgeTransformation by simply adapting the idx_range that is saved to include the periodic point?

In this case, the class has to store index ranges, not only references.

EmilyBourne:
> > I have done half a review, but generally this seems like a very complicated solution to a very simple problem. Could the issue not be fixed in the constructor of EdgeTransformation by simply adapting the idx_range that is saved to include the periodic point?
> 
> In this case, the class has to store index ranges, not only references.
> 

Index ranges are extremely light (a start index and a length) so this is not a problem. Given how small an index range is, it is potentially more efficient to have them stored locally than to use a distant memory block anyway

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -84,43 +227,46 @@ class EdgeTransformation
...
-        }
-        return target_min + rescale_x;
+        static_assert(
+                !std::is_same_v<EdgeDim1, EdgeDim2>,
+                "The two patches have the same conditinous dimension. "
+                "Please specify the input Patch as template parameter "
+                "using .template operator<CurrentPatch>()(current_coord).");
+        return get_equivalent_target_coordinate<
+                CurrentDim,
+                Interface>(current_coord, m_idx_range_patch_1, m_idx_range_patch_2);
```
EmilyBourne, line 244: Why not just give the patch here?
```suggestion
        return get_equivalent_target_coordinate<
                std::conditional_t<std::is_same_v<CurrentDim, EdgeDim1>, Patch1, Patch2>,
                Interface>(current_coord, m_idx_range_patch_1, m_idx_range_patch_2);
```

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -12,6 +12,147 @@
...
+ * @return Equivalent coordinate on the target patch. 
+ */
+template <typename PatchOrDim, class Interface, class CurrentDim>
+Coord<std::conditional_t<
+        std::is_same_v<
+                CurrentDim,
+                typename Interface::Edge1::parallel_grid::continuous_dimension_type>,
+        typename Interface::Edge2::parallel_grid::continuous_dimension_type,
+        typename Interface::Edge1::parallel_grid::continuous_dimension_type>>
+get_equivalent_target_coordinate(
```
EmilyBourne, line 113: I think the use of `PatchOrDim` makes this harder to follow. Can you not ask for the Patch here and find it in `operator()` if you don't have it yet

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -12,6 +12,147 @@
 #include "interface.hpp"
 
 
+/**
+ * @brief Struct to identify the type of an index range on the current 
+ * patch and the index range type on the target patch. 
+ */
+template <class Interface, class PatchOrDim>
+struct IdxRangeCurrentAndTarget;
```
EmilyBourne, line 20: This class with its 4 specialisations seems like a lot of code for something that can be achieved with 2 well placed `conditional_t`

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: It is true that with the previous `static_assert`, it should be possible.

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -12,6 +12,69 @@
...
+                typename Interface::Edge1::parallel_grid::continuous_dimension_type>,
+        class CurrentDim = std::conditional_t<
+                std::is_same_v<CurrentPatch, typename Interface::Edge1::associated_patch>,
+                typename Interface::Edge1::parallel_grid::continuous_dimension_type,
+                typename Interface::Edge2::parallel_grid::continuous_dimension_type>>
+Coord<TargetDim> get_equivalent_target_coordinate(
+        Coord<CurrentDim> const& current_coord,
+        IdxRange<typename Interface::Edge1::parallel_grid> const& idx_range_patch_1,
+        IdxRange<typename Interface::Edge2::parallel_grid> const& idx_range_patch_2)
+{
```
EmilyBourne, line 42: For the readability I would be tempted to do:
```suggestion
template <
        typename CurrentPatch,
        class Interface,
        class CurrentDim>
auto get_equivalent_target_coordinate(
        Coord<CurrentDim> const& current_coord,
        IdxRange<typename Interface::Edge1::parallel_grid> const& idx_range_patch_1,
        IdxRange<typename Interface::Edge2::parallel_grid> const& idx_range_patch_2)
{
    using CurrentEdge = std::is_same_v<CurrentPatch, typename Interface::Edge1::associated_patch>,
                typename Interface::Edge2,
                typename Interface::Edge1>;
    using TargetEdge = Interface::OtherEdge<CurrentEdge>;
    static_assert(std::is_same_v<CurrentDim, typename CurrentEdge::parallel_grid::continuous_dimension_type);
```

EmilyBourne: But this is not much shorter, so feel free to stick with whichever you find easiest to understand

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -223,29 +291,50 @@ class EdgeTransformation
...
         IdxRangeTarget const target_idx_range(combined_idx_range);
 
         int const n_cells_current = current_idx_range.size() - 1;
         int const n_cells_target = target_idx_range.size() - 1;
 
+        // Periodicity property.
+        if constexpr (CurrentGrid::continuous_dimension_type::PERIODIC) {
+            current_idx = (current_idx - current_idx_range.front()).value() < n_cells_current
+                                  ? current_idx
+                                  : IdxCurrent((current_idx - current_idx_range.back()).value());
```
EmilyBourne, line 335: Crappy wifi seems to have lost my comment here. I have 2 comments:
1. Are you sure about the value `IdxCurrent((current_idx - current_idx_range.back()).value())`? I am not sure this works as expected if `current_idx_range.front() != 0`
2. I don't think you need to use integers here. I think it should be (with my correction for point 1 included):
```suggestion
            current_idx = (current_idx < current_idx_range.back()
                                  ? current_idx
                                  : current_idx - current_idx_range.extents());
```

E.g. with `current_idx=11` and `current_idx_range = range(4,10)` we have:
- IdxCurrent((current_idx - current_idx_range.back()).value()) = 11 - 10 = 1
- `current_idx - current_idx_range.extents()` = 11 - 6 = 5

(FYI this is why you are supposed to add/subtract an index step from an index instead of an index as it is harder to make this kind of mistake with steps

src/multipatch/connectivity/edge_transformation.hpp
EmilyBourne: I would prefer a named function which takes a patch index and an `operator()` that calls this function if the patch can be deduced

PaulineVidal: I can add this function. But, I don't think it will replace the operator() templated on the Patch.
Done in https://github.com/gyselax/gyselalibxx/pull/236/commits/6204c981b4d47b5bf36b65c230c224d71f8072a9

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: I didnt know an idx - idxstep could be an idx.

EmilyBourne: > I didnt know an idx - idxstep could be an idx. Or at least I didnt remember.

Hmm, this is kind of the point of an idxstep. It is used to "step" from one index to another. If this was not clear then I guess we need to improve the docs?

src/multipatch/connectivity/edge_transformation.hpp
EmilyBourne, line 42: Why did you choose a public function in the global scope rather than a private method of the class?

PaulineVidal: At the beginning it was because I was using `std::enable_if_t` which could not compile inside of the class because the condition was only depending on template parameters of the class. 
Now, it is true it is not anymore the case. I can replace it inside in private.

src/multipatch/connectivity/edge_transformation.hpp
EmilyBourne: I think you misunderstood my point. My request was for something like:
```cpp
    template <class CurrentPatch>
    Coord<std::conditional_t<std::is_same_v<CurrentPatch, Patch1>, EdgeDim2, EdgeDim1>> transform_edge_coordinate(
            Coord<std::conditional_t<
                    std::is_same_v<CurrentPatch, Patch1>,
                    EdgeDim1,
                    EdgeDim2>> const& current_coord) const
    {
        return get_equivalent_target_coordinate<
                CurrentPatch,
                Interface>(current_coord, m_idx_range_patch_1, m_idx_range_patch_2);
    }

    template <class CurrentDim>
    Coord<std::conditional_t<std::is_same_v<CurrentDim, EdgeDim1>, EdgeDim2, EdgeDim1>> operator()(
            Coord<CurrentDim> const& current_coord) const
    {
        static_assert(
                !std::is_same_v<EdgeDim1, EdgeDim2>,
                "The two patches have the same conditinous dimension. "
                "Please specify the input Patch as template parameter "
                "using .template operator<CurrentPatch>()(current_coord).");
        return get_equivalent_target_coordinate<
                std::conditional_t<std::is_same_v<CurrentDim, EdgeDim1>, Patch1, Patch2>,
                Interface>(current_coord, m_idx_range_patch_1, m_idx_range_patch_2);
    }
```

src/multipatch/connectivity/edge_transformation.hpp
EmilyBourne: image
It looks pretty clear to me. Do you think it is hidden or has it just been a while since you read this part of the docs?

PaulineVidal: I think my mind is just blocked on "a lot of things are not allowed" that I tend to forget what we can do.
It is true, I didnt re-read it for a while. I thought that any operation on Idx will return an IdxStep.

EmilyBourne: I think only Idx-Idx returns an IdxStep (i.e. the distance between the indices)

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: Done in https://github.com/gyselax/gyselalibxx/pull/236/commits/d2db14d448dcd4f478ad2ef01dd913ef9b27882b

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: Now, it is inside the class in private. So I dont think that someone will give the wrong template parameters.
So maybe I prefer the version I wrote.

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -84,43 +86,44 @@ class EdgeTransformation
...
-        }
-        return target_min + rescale_x;
+        static_assert(
+                !std::is_same_v<EdgeDim1, EdgeDim2>,
+                "The two patches have the same conditinous dimension. "
+                "Please specify the input Patch as template parameter "
+                "using .template operator<CurrentPatch>()(current_coord).");
+        using CurrentPatch
+                = std::conditional_t<std::is_same_v<CurrentDim, EdgeDim1>, Patch1, Patch2>;
+        return (*this).get_equivalent_target_coordinate<CurrentPatch>(current_coord);
```
EmilyBourne, line 103:
```suggestion
        return get_equivalent_target_coordinate<CurrentPatch>(current_coord);
```
This is not needed to call operators with friendly names

EmilyBourne, line 103:
```suggestion
        return get_equivalent_target_coordinate<CurrentPatch>(current_coord);
```
This is not needed to call operators with friendly names

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -84,43 +86,44 @@ class EdgeTransformation
...
+    */
+    template <class CurrentPatch>
+    Coord<std::conditional_t<std::is_same_v<CurrentPatch, Patch1>, EdgeDim2, EdgeDim1>>
+    transform_edge_coord(Coord<std::conditional_t<
+                                 std::is_same_v<CurrentPatch, Patch1>,
+                                 EdgeDim1,
+                                 EdgeDim2>> const& current_coord) const
+    {
+        return (*this).get_equivalent_target_coordinate<CurrentPatch>(current_coord);
+    }
```
EmilyBourne: I don't think you need this function and `get_equivalent_target_coordinate` as they are 100% equivalent. You can keep whichever you think has the name that makes the most sense from outside the class

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -153,6 +156,26 @@ class EdgeTransformation
...
+     *      A index on the edge of the current patch.
+     * @tparam CurrentIdx 
+     *      The current index type of the given coordinate index. 
+     * 
+     * @return The analogous coordinate on the target patch. 
+    */
+    template <class CurrentIdx>
+    Coord<std::conditional_t<std::is_same_v<CurrentIdx, IdxEdge1>, EdgeDim2, EdgeDim1>>
+    get_equivalent_coord_from_idx(CurrentIdx const& current_idx) const
+    {
```
EmilyBourne, line 173: Why is this function needed as well as the two above? It seems to me that it is more onerous to maintain 3 docstrings than to duplicate a call to `ddc::coordinate`

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -229,23 +252,39 @@ class EdgeTransformation
...
         IdxRangeTarget const target_idx_range(combined_idx_range);
 
         int const n_cells_current = current_idx_range.size() - 1;
         int const n_cells_target = target_idx_range.size() - 1;
 
+        // Periodicity property.
+        if constexpr (CurrentGrid::continuous_dimension_type::PERIODIC) {
+            current_idx = (current_idx < current_idx_range.back())
+                                  ? current_idx
+                                  : current_idx - current_idx_range.extents() + 1;
```
EmilyBourne, line 285: Why do you need the `+1` here?

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -317,7 +407,16 @@ class EdgeTransformation
...
+            CurrentCoord target_equivalent_coord_mid
+                    = get_equivalent_coord_from_idx(target_idx_mid);
+
+            // Periodicity property.
+            if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
+                if (target_idx_mid == target_idx_range.back()) {
+                    target_equivalent_coord_mid
+                            = get_equivalent_coord_from_idx(target_idx_range.front());
+                }
+            }
```
EmilyBourne, line 419: Is this needed? Shouldn't the periodicity be handled inside `get_equivalent_coord_from_idx`, i.e. before you get here?

tests/multipatch/connectivity/index_transformation_non_uniform_periodic.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    Patch2::IdxRange1 const idx_range_r2;
+    Patch2::IdxRange2 const idx_range_theta2;
+
+public:
+    IndexTransformationNonUniformPeriodicTest()
+        : idx_range_r1(Patch1::Idx1(0), r1_npoints)
+        , idx_range_theta1(Patch1::Idx2(0), theta1_npoints)
+        , idx_range_r2(Patch2::Idx1(0), r2_npoints)
+        , idx_range_theta2(Patch2::Idx2(0), theta2_npoints) {};
```
EmilyBourne, line 47:
```suggestion
        , idx_range_theta2(Patch2::Idx2(0), theta2_npoints) {}
```

tests/multipatch/connectivity/coord_transformation_periodic.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    Patch2::IdxRange1 const idx_range_r2;
+    Patch2::IdxRange2 const idx_range_theta2;
+
+public:
+    CoordinateTransformationPeriodicTest()
+        : idx_range_r1(Patch1::Idx1(0), r1_npoints)
+        , idx_range_theta1(Patch1::Idx2(0), theta1_npoints)
+        , idx_range_r2(Patch2::Idx1(0), r2_npoints)
+        , idx_range_theta2(Patch2::Idx2(0), theta2_npoints) {};
```
EmilyBourne, line 46:
```suggestion
        , idx_range_theta2(Patch2::Idx2(0), theta2_npoints) {}
```

tests/multipatch/connectivity/index_transformation_uniform_periodic.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    Patch2::IdxRange1 const idx_range_r2;
+    Patch2::IdxRange2 const idx_range_theta2;
+
+public:
+    IndexTransformationUniformPeriodicTest()
+        : idx_range_r1(Patch1::Idx1(0), r1_npoints)
+        , idx_range_theta1(Patch1::Idx2(0), theta1_npoints)
+        , idx_range_r2(Patch2::Idx1(0), r2_npoints)
+        , idx_range_theta2(Patch2::Idx2(1), theta2_npoints) {};
```
EmilyBourne, line 46:
```suggestion
        , idx_range_theta2(Patch2::Idx2(1), theta2_npoints) {}
```

EmilyBourne (changes requested):

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: Oh yes, true I didnt notice.

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: I removed it.

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: It comes from a misunderstanding of your previous comment.
This function can deduce the patch from the index. It is possible that we write a lot
```cpp
TargetCoord coord = transform_edge_coord<CurrentPatch>(ddc::coordinate(idx));
```
so I wonder it is worth it to keep this function to write
```cpp
TargetCoord coord = get_equivalent_coord_from_idx(idx);
```

EmilyBourne: My feeling would be that it is not worth it.
I don't think there is a readability gain here so the only argument for this seems to be that it is shorter.
You save 22 characters per call but add 810 characters for the definition (including the docstring, 403 without), so you need to call this function at least 36 times to gain anything. I don't think you will use it this extensively

src/multipatch/connectivity/edge_transformation.hpp
EmilyBourne: >  but then it may causes errors if for example
> current_idx = current_idx_range.back() +current_idx_range.back() + 1

I think this is significantly less likely than:

> current_idx = current_idx_range.back() + 1

so I like the version with `<`.

If I have correctly understood the purpose of the `+1` maybe this would be clearer as:
```cpp
        if constexpr (CurrentGrid::continuous_dimension_type::PERIODIC) {
            current_idx = (current_idx < current_idx_range.back())
                                  ? current_idx
                                  : current_idx - n_cells_current;
```
?

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: If it is a question of number of lines.
Done in https://github.com/gyselax/gyselalibxx/pull/236/commits/da80eb333d350cc144d872979e1ff148df577619#diff-6ca366f244ec8fff8736b57bfc1bb1f221c1388e348bd154e2b7b484073586d0
and https://github.com/gyselax/gyselalibxx/pull/236/commits/848bf7e3977124bc8f7fe4c1503990bcbb4dc806

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: Actually it would work.

PaulineVidal: Changed in https://github.com/gyselax/gyselalibxx/pull/236/commits/5546142148e783fcbf59ff623578f6ef03330196

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -84,23 +86,53 @@ class EdgeTransformation
...
     template <class CurrentDim>
     Coord<std::conditional_t<std::is_same_v<CurrentDim, EdgeDim1>, EdgeDim2, EdgeDim1>> operator()(
             Coord<CurrentDim> const& current_coord) const
     {
-        // The other continuous dimension
+        static_assert(
+                !std::is_same_v<EdgeDim1, EdgeDim2>,
+                "The two patches have the same conditinous dimension. "
+                "Please specify the input Patch as template parameter "
+                "using .template operator<CurrentPatch>()(current_coord).");
```
EmilyBourne, line 100:
```suggestion
                "using .template transform_edge_coord<CurrentPatch>()(current_coord).");
```

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -110,14 +142,18 @@ class EdgeTransformation
         double const current_length = ddcHelper::total_interval_length(current_idx_range);
 
         Coord<TargetDim> const target_min = ddc::coordinate(target_idx_range.front());
-        double const target_length = ddcHelper::total_interval_length(target_idx_range);
+        double target_length = ddcHelper::total_interval_length(target_idx_range);
+
```
EmilyBourne, line 146: Why this change?

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -223,29 +259,58 @@ class EdgeTransformation
...
+            }
+            counter = 0;
+            while ((current_idx >= current_idx_range.back()) && (counter < 10)) {
+                counter++;
+                current_idx -= n_cells_current;
+            }
+        } else {
+            assert(current_idx <= current_idx_range.back());
+            assert(current_idx >= current_idx_range.front());
+        }
```
EmilyBourne, line 312:
```suggestion
        }
        assert(current_idx <= current_idx_range.back());
        assert(current_idx >= current_idx_range.front());
```
If you want to be completely general I guess these assertions shouldn't be in the `else` condition as you have no guarantee that the periodicity returns to the correct range (due to the n_iter_max = 10)

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: std::fmod cannot work with const variable 
https://github.com/gyselax/gyselalibxx/pull/236/files#diff-6ca366f244ec8fff8736b57bfc1bb1f221c1388e348bd154e2b7b484073586d0R155

EmilyBourne: Are you sure? Nothing in the prototype indicates this:
image

and you are saving the result to `rescale_x` not `target_length`

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -223,29 +259,58 @@ class EdgeTransformation
...
+            std::size_t counter = 0;
+            while ((current_idx < current_idx_range.front()) && (counter < 10)) {
+                counter++;
+                current_idx += n_cells_current;
+            }
+            counter = 0;
+            while ((current_idx >= current_idx_range.back()) && (counter < 10)) {
+                counter++;
+                current_idx -= n_cells_current;
+            }
```
EmilyBourne, line 308: Why is this function the only one where the index may be so far away from the domain of interest?
E.g. it is not a problem here:
https://github.com/gyselax/gyselalibxx/pull/236/files#diff-6ca366f244ec8fff8736b57bfc1bb1f221c1388e348bd154e2b7b484073586d0R342-R344
https://github.com/gyselax/gyselalibxx/pull/236/files#diff-6ca366f244ec8fff8736b57bfc1bb1f221c1388e348bd154e2b7b484073586d0R412-R414

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: I can recheck but I had an error.

PaulineVidal: Now it seems it is working.

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: Because `current_idx` is given by the user. `Target_idx` is computed here. It cannot be outside of the index range.

tests/multipatch/connectivity/index_transformation_non_uniform_periodic.cpp
```diff
@@ -0,0 +1,177 @@
...
+    using EdgeR2F = Edge<Patch2, GridR<2>, FRONT>;
+    using Interface12 = Interface<EdgeR1B, EdgeR2F, false>;
+
+    // Coordinate transformation .................................................................
+    EdgeTransformation<Interface12> index_transformation(idx_range_theta1, idx_range_theta2);
+
+    Patch1::Idx2 test_idx_theta1(6);
+    Patch2::Idx2 test_idx_theta2(index_transformation(test_idx_theta1));
+
+    EXPECT_EQ((test_idx_theta2 - idx_range_theta2.front()).value(), 3);
```
EmilyBourne, line 126:
```suggestion
    EXPECT_EQ(test_idx_theta2, Patch2::Idx2(3));
```

EmilyBourne: Same below

src/multipatch/connectivity/edge_transformation.hpp
EmilyBourne: Ok. Maybe a comment would be useful?

tests/multipatch/connectivity/index_transformation_non_uniform_periodic.cpp
PaulineVidal: Done in https://github.com/gyselax/gyselalibxx/pull/236/commits/1ac2563bfe7ba83a8f221230f4ed4ea7d80029c5

EmilyBourne: I guess some of your index systems don't start at 0 hence the
```cpp
EXPECT_EQ(test_idx_y3, Patch3::Idx2( 3+3));
```
?
But why do you have:
```cpp
EXPECT_EQ(test_idx_theta2 , Patch2::Idx2(0+1));
```
?

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: Where do you want the comment? Why this treatment of current_idx ? or why not for target_idx ?

PaulineVidal: It is a bit weird to say, "our computation computes the right index, so _inside_ the domain". But, I can add assert if you prefer? https://github.com/gyselax/gyselalibxx/pull/236/commits/168a2d6fe53925b697c81641e1ca44843cb3f9f5

tests/multipatch/connectivity/index_transformation_non_uniform_periodic.cpp
EmilyBourne: Ah ok I assumed this was front+idx_step not idx_step+front. Maybe a comment would be useful then as it seems it is not obvious (sorry)

src/multipatch/connectivity/edge_transformation.hpp
```diff
@@ -265,20 +328,28 @@ class EdgeTransformation
...
                 }
+
+                // Apply periodicity property of the domain.
+                if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
+                    if (target_idx == target_idx_range.back()) {
+                        target_idx = target_idx_range.front();
+                    }
+                }
+                assert(target_idx <= target_idx_range.back());
+                assert(target_idx >= target_idx_range.front());
```
EmilyBourne, line 345:
```suggestion
                assert(target_idx <= target_idx_range.back());
                assert(target_idx >= target_idx_range.front());
                // Apply periodicity property of the domain.
                if constexpr (TargetGrid::continuous_dimension_type::PERIODIC) {
                    if (target_idx == target_idx_range.back()) {
                        target_idx = target_idx_range.front();
                    }
                }
```
?

src/multipatch/connectivity/edge_transformation.hpp
EmilyBourne: Assertions seem like a good idea to indicate what you are expecting :+1:

Otherwise I would have just put a comment like:
```
// target_idx is calculated so it is inside target_idx_range
```
or maybe better:
```
// current_idx is provided by the user so it may be outside the expected range
```

src/multipatch/connectivity/edge_transformation.hpp
PaulineVidal: So

assert(target_idx < target_idx_range.back()); just after the 

if (target_idx == target_idx_range.back()) {

                        target_idx = target_idx_range.front();

} ?
I think it is a bit too much.
I prefer to put the assert before then

PaulineVidal: https://github.com/gyselax/gyselalibxx/pull/236/commits/76c81d196ed8e40ea5494d8f8e88a688d263f384

tests/multipatch/connectivity/index_transformation_non_uniform_periodic.cpp
PaulineVidal: ok

PaulineVidal: I changed it here: https://github.com/gyselax/gyselalibxx/pull/236/commits/0f032199a102e6af564d31f4fdc6bfa2e024c30c

EmilyBourne (approved):

PR #245, opened 2025-04-08
Add a fence before communication

tpadioleau:
Correct me if I am wrong, I don't think we promise a that `rho_local` is ready after calling `m_local_charge_density_calculator`. This can lead to a race condition between an asynchronous backend and the MPI communication. I think we need to add a fence before `MPI_Allreduce`.

Note that it should not affect the benchmark results done with Kokkos Tools as regions will block.


EmilyBourne (approved):
I think you are right as this seems to be a GPU aware MPI call

PR #259, opened 2025-04-21
spack.yaml: add ginkgo and fftw+float

freifrauvonbleifrei:
helped me to install gyselalibxx with spack!

### DIFF
--- a/spack.yaml
+++ b/spack.yaml
@@ -5,6 +5,8 @@ spack:
   - 'cmake@3.15:'
   - 'googletest +gmock'
   - 'fftw ~mpi precision=double'
+  - 'fftw ~mpi precision=float'
+  - 'ginkgo'
   - 'lapack'
   - 'paraconf@0.4.15:'
   - 'pdi'


tpadioleau:
Thank you @freifrauvonbleifrei for this update. This file has been there for a while and is now kind of deprecated. We will update soon the process to install a working CPU environment.

spack.yaml
```diff
@@ -5,6 +5,8 @@ spack:
   - 'cmake@3.15:'
   - 'googletest +gmock'
   - 'fftw ~mpi precision=double'
+  - 'fftw ~mpi precision=float'
```
tpadioleau, line 8: I am surprised that the two fftw lines work fine together :/

tpadioleau: Yes this is what we have in https://github.com/gyselax/gyselalibxx/pull/260.

I guess this led to install different `.so` and somehow worked fine with `FindFFTW.cmake` but definitely not what I recommend to do.

freifrauvonbleifrei:
OK, I think I'll close this in favor of #260 ! thx!

tpadioleau:
> OK, I think I'll close this in favor of #260 ! thx!

Thank you, I hope we will get a simpler starting guide in #260.

PR #272, opened 2025-04-30
Add x0 and y0 parameters to circular and czarny coordinate change operators

### LINKED ISSUE #271: [FEATURE]: Add x0 and y0 parameters to CurvilinearToCartesian mappings
### Mathematical operator

Add x0 and y0 parameters to CurvilinearToCartesian mappings.

### Implementation Ideas

E.g. a circular mapping should not be:
```math
x(r,\theta) = r \cos(\theta) \quad
y(r,\theta) = r \sin(\theta)
```
but rather:
```math
x(r,\theta) = r \cos(\theta) + x_0 \quad
y(r,\theta) = r \sin(\theta) + y_0
```

### Benefits

This allows a toroidal coordinate system to be correctly defined

### Contribution

- [x] I'm willing to work on this issue

EmilyBourne:
Add x0 and y0 parameters to circular and czarny coordinate change operators. Fixes #271

### DIFF
--- a/src/mapping/cartesian_to_circular.hpp
+++ b/src/mapping/cartesian_to_circular.hpp
@@ -68,6 +68,12 @@ private:
     double m_y0;
 
 public:
+    /**
+     * @brief Instantiate a CartesianToCircular from parameters.
+     *
+     * @param[in] x0 The x-coordinate of the centre of the circle (0 by default).
+     * @param[in] y0 The y-coordinate of the centre of the circle (0 by default).
+     */
     CartesianToCircular(double x0 = 0.0, double y0 = 0.0) : m_x0(x0), m_y0(y0) {}
 
     /**
--- a/src/mapping/cartesian_to_circular.hpp
+++ b/src/mapping/cartesian_to_circular.hpp
@@ -17,23 +17,23 @@ class CircularToCartesian;
  *
  * The mapping @f$ (x,y)\mapsto (r,\theta) @f$ is defined as follow :
  *
- * @f$ r(x,y) = \sqrt x^2+y^2 ,@f$
+ * @f$ r(x,y) = \sqrt (x-x_0)^2+(y-y_0)^2 ,@f$
  *
- * @f$ \theta(x,y) = atan2(\frac{y}{x}) .@f$
+ * @f$ \theta(x,y) = atan2(\frac{y-y_0}{x-x_0}) .@f$
  *
  * It and its Jacobian matrix are invertible everywhere except for @f$ r = 0 @f$.
  *
  * The Jacobian matrix coefficients are defined as follow
  *
- * @f$ J_{11}(x,y)  =\frac{x}{\sqrt{x^2+y^2}}  @f$
+ * @f$ J_{11}(x,y)  =\frac{x-x_0}{\sqrt{(x-x_0)^2+(y-y_0)^2}}  @f$
  *
- * @f$ J_{12}(x,y)  =\frac{y}{\sqrt{x^2+y^2}}  @f$
+ * @f$ J_{12}(x,y)  =\frac{y-y_0}{\sqrt{(x-x_0)^2+(y-y_0)^2}}  @f$
  *
- * @f$ J_{21}(x,y)  =\frac{-y}{x^2+y^2}  @f$
+ * @f$ J_{21}(x,y)  =\frac{-(y-y_0)}{(x-x_0)^2+(y-y_0)^2}  @f$
  *
- * @f$ J_{22}(x,y)  =\frac{x}{x^2+y^2}  @f$
+ * @f$ J_{22}(x,y)  =\frac{x-x_0}{(x-x_0)^2+(y-y_0)^2}  @f$
  *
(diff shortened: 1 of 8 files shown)


EmilyBourne:
The Poisson failure only occurs when `m_x0` or `m_y0` is not an integer (e.g `6.1` instead of `6.0`). This implies an erroneous integer conversion buried in the code

EmilyBourne:
The matrix is the same. This probably implies that the problem is with the RHS

EmilyBourne:
The problem is with the manufactured solution which was not updated. The cartesian solution includes a dependency on x/y which was not updated to include x0/y0

src/mapping/cartesian_to_circular.hpp
```diff
@@ -63,16 +63,26 @@ class CartesianToCircular
...
+    explicit CartesianToCircular(double x0 = 0.0, double y0 = 0.0) : m_x0(x0), m_y0(y0) {}
 
     /**
      * @brief Instantiate a CartesianToCircular from another CartesianToCircular (lvalue).
      *
      * @param[in] other
      * 		CartesianToCircular mapping used to instantiate the new one.
      */
-    KOKKOS_FUNCTION CartesianToCircular(CartesianToCircular const& other) {}
+    KOKKOS_DEFAULTED_FUNCTION CartesianToCircular(CartesianToCircular const& other) = default;
```
gdgirard, line 85: Can you explain again the goal of this change ?

EmilyBourne: I'm not sure why we were using:
```cpp
KOKKOS_FUNCTION CartesianToCircular(CartesianToCircular const& other) {}
```
previously as this would be the default implementation of the copy constructor. Now there are 2 extra parameters I can either explicitly copy them:
```cpp
KOKKOS_FUNCTION CartesianToCircular(CartesianToCircular const& other)
    : m_x0(x0)
    , m_y0(y0)
{}
```
or use the default copy constructor which does exactly the same thing.
I prefer the method with the default copy constructor as it is more compact.

tests/geometryRTheta/advection_rtheta/advection_all_tests.cpp
```diff
@@ -347,11 +347,11 @@ int main(int argc, char** argv)
 
 
     // SET THE DIFFERENT PARAMETERS OF THE TESTS ------------------------------------------------
-    CircularToCartMapping const from_circ_map;
-    CircularToPseudoCartMapping const to_pseudo_circ_map;
-    CartesianToCircular<X, Y, R, Theta> to_circ_map;
-    CzarnyToCartMapping const from_czarny_map(0.3, 1.4);
-    CartesianToCzarny<X, Y, R, Theta> const to_czarny_map(0.3, 1.4);
+    CircularToCartMapping const from_circ_map(6.2);
```
gdgirard, line 350: Is 6.2 an arbitrary choice ?
Not sure to understand but an I wrong if I understand that this line is equivalent to 
`CircularToCartMapping const from_circ_map(x_0=6.2, y_0=0); ?`

EmilyBourne: Yes. I wanted to pick something that was not 0 or 1 to ensure that the value is checked properly (6.2m is the major radius of ITER).

Yes this line is equivalent to:
> `CircularToCartMapping const from_circ_map(x_0=6.2, y_0=0); `

gdgirard: But for a unitary test why you don't choose something completely arbitrary with a non null-value for y_0 to test also that all is correct in `y` direction ?

gdgirard: More general questions when you have several default value as input of a function the deduction is made with the order of the parameter ? If yes, I would be more in favor of writing explicitly the name of the variable. Is it not a proper way to do in C++ ?

src/mapping/cartesian_to_circular.hpp
gdgirard: OK, thank you for the clarification.

tests/geometryRTheta/advection_rtheta/advection_all_tests.cpp
EmilyBourne: > But for a unitary test why you don't choose something completely arbitrary with a non null-value for y_0 to test also that all is correct in y direction ?

I will modify this :+1:


> More general questions when you have several default value as input of a function the deduction is made with the order of the parameter ?

Yes.

> If yes, I would be more in favor of writing explicitly the name of the variable.

Unfortunately there is no syntax for this in C++ so you cannot get it to compile.

EmilyBourne: Done : [f20268c](https://github.com/gyselax/gyselalibxx/pull/272/commits/f20268ca7ede9a0b83419cca7a6391de775ce475)

gdgirard (approved):

PR #275, opened 2025-05-02
Add a compilation with clang

### LINKED ISSUE #29: Add a compilation with clang
A clang compilation test in the CI would be beneficial for multiple reasons:
1. It would test compilation for Adastra
3. It would catch missing `KOKKOS_LAMBDA` errors (these are only warnings with a Nvidia compiler)

EmilyBourne:
Add a compilation with clang to the CI. Fixes #29 . Fix compilation errors.

To merge after #280

### DIFF
--- a/.github/workflows/cpu_tests.yml
+++ b/.github/workflows/cpu_tests.yml
@@ -103,20 +103,20 @@ jobs:
       CXX: ${{ matrix.compiler == 'gnu' && 'g++' || 'clang++' }}
     steps:
       - name: Checkout gyselalibxx
-        if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
+        #if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
         uses: actions/checkout@v4
         with:
           submodules: recursive
       - name: "Filter tests"
-        if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
+        #if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
         uses: ./.github/actions/test_filter
       - name: Build code
-        if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
+        #if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
         uses: ./.github/actions/build_code
         with:
           toolchain: toolchains/docker.gyselalibxx_env/${{ matrix.toolchain }}.cmake
       - name: Run tests
-        if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
+        #if: ${{ needs.pre_job.outputs.should_skip == 'false' }}
         uses: ./.github/actions/run_tests
       - name: Check Test Coverage
         if: needs.pre_job.outputs.should_skip == 'false' && matrix.toolchain == 'tests_coverage_toolchain'
--- a/.github/workflows/cpu_tests.yml
+++ b/.github/workflows/cpu_tests.yml
@@ -74,7 +74,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       image_tag: ${{ steps.docker.outputs.image_tag }}
-    if: github.repository == 'gyselax/gyselalibxx' && (github.event_name != 'pull_request' || github.event.pull_request.draft == false)
+    #if: github.repository == 'gyselax/gyselalibxx' && (github.event_name != 'pull_request' || github.event.pull_request.draft == false)
     steps:
       - uses: actions/checkout@v4
       - id: docker
--- a/.github/workflows/cpu_tests.yml
+++ b/.github/workflows/cpu_tests.yml
@@ -94,7 +94,7 @@ jobs:
       options: --user root
     name: CPU Test (${{ matrix.toolchain }}, ${{ matrix.compiler }})
     # Don't run coverage for clang (too slow)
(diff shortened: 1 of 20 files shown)


EmilyBourne:
cc @PaulineVidal 
The failing tests are in the multipatch onion handling.
Do you have any idea why this result might be compiler dependent? Could this be rounding errors?

PaulineVidal:
> Could this be rounding errors?

Yes, I think the two tests failing are the one testing that the coordinate on the boundary are considered as in the patch. 
* This test https://github.com/gyselax/gyselalibxx/actions/runs/14798744130/job/41554654842?pr=275#step:7:873 checks that the coordinate (1.5,0) located on the edge of the Patch2 and mapped to the physical domain is considered as on the Patch2. 

* The test https://github.com/gyselax/gyselalibxx/actions/runs/14798744130/job/41554654842?pr=275#step:7:848, I suspect that it fails for the same reason on GPU.

EmilyBourne:
> > Could this be rounding errors?
> 
> Yes, I think the two tests failing are the one testing that the coordinate on the boundary are considered as in the patch.
> 
> * This test https://github.com/gyselax/gyselalibxx/actions/runs/14798744130/job/41554654842?pr=275#step:7:873 checks that the coordinate (1.5,0) located on the edge of the Patch2 and mapped to the physical domain is considered as on the Patch2.
> * The test https://github.com/gyselax/gyselalibxx/actions/runs/14798744130/job/41554654842?pr=275#step:7:848, I suspect that it fails for the same reason on GPU.
> 
> In both case, it may returns "outside" of the domain (instead of "Patch2") if there are rounding errors. It was not the case when it was merged, but maybe with another compiler, the error can be different.

Ah yes I see that you hard-code the values here:
https://github.com/gyselax/gyselalibxx/blob/67a30189c3a1d4e2662416acb4d84113b80dd02a/tests/multipatch/connectivity/patch_locator_onion_shape_2patches.cpp#L207-L213

If some of those values are designed to be on the patch boundaries then this test will be very susceptible to rounding errors. This should be fixed but it does not need to be done in this MR. As I am on holiday next week I will open an issue and mark this PR as blocked

EmilyBourne:
Blocked by #276

tpadioleau (approved):
LGTM

.github/workflows/cpu_tests.yml
```diff
@@ -85,13 +85,25 @@ jobs:
...
     strategy:
-      matrix: ${{fromJson(needs.pre_job.outputs.matrix)}}
+      matrix:
+        toolchain: ${{fromJson(needs.pre_job.outputs.toolchain_matrix)}}
+        compiler: ['gnu', 'clang']
       fail-fast: false
     container:
       image: ${{ needs.docker_update.outputs.image_tag }}
       options: --user root
     name: CPU Test (${{ matrix.toolchain }})
```
EmilyBourne, line 95:
```suggestion
     name: CPU Test (${{ matrix.toolchain }}, ${{ matrix.toolchain }})
```

docker/gyselalibxx_env/Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN chmod +x /bin/bash_run \
  && apt-get update -y \
  && apt-get install -y --no-install-recommends \
     build-essential \
+    clang \
```
tpadioleau, line 28:
```suggestion
    clang \
    libomp-dev \
```

.github/workflows/cpu_tests.yml
```diff
@@ -85,12 +85,20 @@ jobs:
     runs-on: ubuntu-latest
     needs: [pre_job, docker_update]
     strategy:
-      matrix: ${{fromJson(needs.pre_job.outputs.matrix)}}
+      matrix:
+        toolchain: ${{fromJson(needs.pre_job.outputs.toolchain_matrix)}}
+        compiler: ['gnu', 'clang']
```
tpadioleau, line 90: Small suggestion

```suggestion
        compilers:
          - c: gcc
            cxx: g++
          - c: clang
            cxx: clang++
```

tpadioleau (approved):
lgtm

.github/workflows/cpu_tests.yml
EmilyBourne: I didn't realise you could create sub-sets like this. So the env command below would become:
```
    env:
      CC: ${{ matrix.compiler.c }}
      CXX: ${{ matrix.compiler.cxx }}
```
?

EmilyBourne: image
Are you sure this is possible?

tpadioleau: Maybe I did a mistake, we have something like this in DDC https://github.com/CExA-project/ddc/blob/35de9ca101f1639b0fa9d516ad7aa3363566a9fc/.github/workflows/tests-ubuntu.yaml#L201-L224

EmilyBourne: Strange the editor highlights it as buggy but it doesn't seem to actually cause any problems:
[154e8f5](https://github.com/gyselax/gyselalibxx/pull/275/commits/154e8f5816863a285ba9b66c7364874d15701a4f)

tpadioleau: Indeed this is the first time I also notice this. I remember Paul Z. telling me the support is not really official.

PR #277, opened 2025-05-12
Rename fdistribu -> density in r,theta geometry

### LINKED ISSUE #148: Rename fdistribu -> density in geometryRTheta
A distribution function is defined on velocity dimensions but the (r, theta) geometry has no velocity dimensions. In this case what is advected is not a distribution function but rather a density. The naming conventions in this part of the code should be updated to reflect that.

Remark: This work should be done **after** the generalisation of reusable sections of this geometry. This will ensure that we don't need to change the names back later.

EmilyBourne:
Rename `fdistribu` which indicates a distribution function (i.e. a function defined in phase space so on spatial **and** velocity dimensions) and `allfdistribu` which indicates a distribution function defined for all the different species to `density` in the r, theta geometry. This is done because the function of interest in the r, theta geometry has no dependency on velocity space or on species. Fixes #148

### DIFF
--- a/src/geometryRTheta/poisson/iqnsolver.hpp
+++ b/src/geometryRTheta/poisson/iqnsolver.hpp
@@ -22,11 +22,11 @@ public:
      *      The solution of the Quasi-Neutrality equation.
      * @param[out] electric_field
      *      The electric field @f$E = -\nabla \phi@f$.
-     * @param[in] allfdistribu
+     * @param[in] density
      *      The rhs of the Quasi-Neutrality equation.
      */
     virtual void operator()(
             host_t<DFieldRTheta> electrostatic_potential,
             host_t<DVectorFieldRTheta<X, Y>> electric_field,
-            host_t<DConstFieldRTheta> allfdistribu) const = 0;
+            host_t<DConstFieldRTheta> density) const = 0;
 };
--- a/src/geometryRTheta/time_solver/bsl_predcorr.hpp
+++ b/src/geometryRTheta/time_solver/bsl_predcorr.hpp
@@ -114,7 +114,7 @@ public:
     }
 
     host_t<DFieldRTheta> operator()(
-            host_t<DFieldRTheta> allfdistribu_host,
+            host_t<DFieldRTheta> density_host,
             double const dt,
             int const steps) const override
     {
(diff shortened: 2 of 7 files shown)


EmilyBourne:
@PaulineVidal do you know if it is expected that so little of the bsl predcorr operators are tested? Are they only used in simulations ?

PaulineVidal:
> @PaulineVidal do you know if it is expected that so little of the bsl predcorr operators are tested? Are they only used in simulations ?

Yes, we agreed that the simulations are the tests for the predictor corrector. As a test for a predictor corrector would be very similar to the simulation a diocotron (or vortex merger) which typical example models.

EmilyBourne:
> > @PaulineVidal do you know if it is expected that so little of the bsl predcorr operators are tested? Are they only used in simulations ?
> 
> Yes, we agreed that the simulations are the tests for the predictor corrector. As a test for a predictor corrector would be very similar to the simulation a diocotron (or vortex merger) which typical example models.

Ok. Fine for this PR, but I would prefer some kind of explicit test to prevent regression. Even if it is a non-regression test. As far as I am aware, the diocotron/vortex merger simulations don't have analytical solutions so it is not trivial to detect when the results change in a problematic way. If we had some kind of analytical solution we could do something similar to what is done with the Landau damping for the X, Vx geometry (where we check the growth rate and frequency to ensure that they still follow the theoretical predictions)

PaulineVidal:
> diocotron/vortex merger simulations don't have analytical solutions so it is not trivial to detect when the results change in a problematic way. If we had some kind of analytical solution we could do something similar to what is done with the Landau damping for the X, Vx geometry (

For the guiding-center model, I don't think we have analytical solutions. But we can add a test on the growth rate. The computation of the growth rate is made in the post-process in a python file. Either we do as in the advection test or polar Poisson test where the test is from a python test, either we compute the integrals in the Cpp code. 
I dont know in this case, if we duplicate to have a test _and_ a simulation pretty similar. If we shift the simulation in the tests folder. Or if we had Google tests in the simulation folder.

EmilyBourne:
> For the guiding-center model, I don't think we have analytical solutions. But we can add a test on the growth rate.

Do you have a theoretical solution for the growth rate?

> The computation of the growth rate is made in the post-process in a python file. Either we do as in the advection test or polar Poisson test where the test is from a python test, either we compute the integrals in the Cpp code. I dont know in this case, if we duplicate to have a test _and_ a simulation pretty similar. If we shift the simulation in the tests folder. Or if we had Google tests in the simulation folder.

The simulations are always compiled anyway. For Landau and Bump On Tail, we have a script in the tests folder which calls the simulation. E.g. see:
https://github.com/gyselax/gyselalibxx/tree/devel/tests/geometryXVx/landau

EmilyBourne:
I have created an issue. I don't consider this to be blocking for this PR which is just a simple renaming.

PaulineVidal (approved):

PR #282, opened 2025-05-13
Remove unused Eigen dependency

### LINKED ISSUE #279: [Refactor]: Remove Eigen dependency
### Current Implementation

Eigen was used in pdepoissonsolver.hpp but has since been removed (replaced by ginkgo). It is still mentioned in the dependencies (including the docker and submodule dependencies) though.

### Suggested Improvement

The unnecessary dependencies should be removed.

### Benefits

This will simplify the installation

EmilyBourne:
Fixes #279


tpadioleau (approved):
ok for me

PR #286, opened 2025-05-14
Allow overloaded functions in coordinate convertor classes

### LINKED ISSUE #285: [Bug]: Multiple jacobian functions in mapping class
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

Add possibility to create several jacobian functions in the same mapping class.
This will be mandatory for ToroidalToCartesian mapping when we will move to 3D space.

### Version of the code used

f7f5e24900e0e4f902298813a08736a11b5ab322

### Steps to Reproduce

In Gysela-X I would like to be able to call:
`MetricTensorEvaluator<ToroidalToCartesianMapping, Coord<Rho, Theta, Phi>> get_metric(mapping);`

### Environment

_No response_

### Logs or Screenshots

_No response_

### How does this bug impact your work?

Minor - It’s an inconvenience, but I can work around it.

EmilyBourne:
Allow overloaded functions in classes in `mapping/`. These classes are often tested with `is_mapping_v`, `has_jacobian_v` etc. These static checkers are designed to check if a specified method exists, but before this PR they did not handle overloaded methods correctly. No method was identified even if the required method existed. This PR fixes this by changing the implementation of `CheckClassAttributeExistence` to use the result type of the method (selected using the arguments) instead of the address of the method. Fixes #285.

### DIFF
--- a/src/mapping/circular_to_cartesian.hpp
+++ b/src/mapping/circular_to_cartesian.hpp
@@ -151,6 +151,15 @@ public:
         return r;
     }
 
+    /**
+     * @brief Compute the Jacobian, the determinant of the Jacobian matrix of the mapping
+     * as a function of the radius.
+     *
+     * @param[in] coord
+     *          The radial coordinate where we evaluate the Jacobian.
+     *
+     * @return A double with the value of the determinant of the Jacobian matrix.
+     */
     KOKKOS_INLINE_FUNCTION double jacobian(Coord<R> const& coord) const
     {
         return ddc::get<R>(coord);
--- a/src/mapping/circular_to_cartesian.hpp
+++ b/src/mapping/circular_to_cartesian.hpp
@@ -151,6 +151,11 @@ public:
         return r;
     }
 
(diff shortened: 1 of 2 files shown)


src/mapping/circular_to_cartesian.hpp
```diff
@@ -151,6 +151,11 @@ class CircularToCartesian
         return r;
     }
 
+    KOKKOS_INLINE_FUNCTION double jacobian(Coord<R> const& coord) const
```
EmilyBourne, line 154:
```suggestion

    /**
     * @brief Compute the Jacobian, the determinant of the Jacobian matrix of the mapping
     * as a function of the radius.
     *
     * @param[in] coord
     *          The radial coordinate where we evaluate the Jacobian.
     *
     * @return A double with the value of the determinant of the Jacobian matrix.
     */
    KOKKOS_INLINE_FUNCTION double jacobian(Coord<R> const& coord) const
```

tpadioleau (approved):

PR #294, opened 2025-05-15
Add a ConstantPartialDerivative calculator

### LINKED ISSUE #291: [FEATURE]: Add a NullPartialDerivativeCalculator
### Mathematical operator

$$\partial_X f = 0$$

### Implementation Ideas

An extra sub-class of `IPartialDerivativeCalculator` can be created

### Benefits

This will simplify the axi-symmetric case which needs to juggle between 2D objects (in r, theta) and 3D objects (in r, theta, phi, e.g. B)

### Contribution

- [ ] I'm willing to work on this issue

EmilyBourne:
Add a ConstantPartialDerivative calculator to help with axi-symmetric simulations. Fixes #291


tests/math_tools/test_partial_derivatives.cpp
```diff
@@ -735,4 +799,13 @@ TEST(PartialDerivative, CentralFDMPartialDerivativeWithBV)
 
     EXPECT_LE(relative_error_order_y, TOL);
 }
+
+TEST(PartialDerivative, ConstantPartialDerivative)
+{
+    const double val(3.5);
+    PartialDerivativeTestConstant<10, 10> const test(val);
```
gdgirard, line 806: Would be maybe better to test with a different number of cells in x and y ?

EmilyBourne: Not sure it makes much difference for this particular operator but I agree in general
```suggestion
    PartialDerivativeTestConstant<10, 5> const test(val);
```

gdgirard: I completely agree with you. It is not specific to this test but I think it would be better to take the habit to choose different values for all directions for tests.

PR #402, opened 2025-08-05
Add a 9 patch non uniform geometry for the periodic strips

PaulineVidal:
Currently,  the 9 patch periodic strip geometry exists only for uniform grids. 
Some test needs to have it on non-uniform grids. 
In this branch, a `9patches_2d_periodic_strips.hpp` files is created. It is inspired from the `3patches_2d_non_periodic.hpp` file to avoid duplication. `9patches_2d_periodic_strips_uniform.hpp`  is modified to rely on the new file. And `9patches_2d_periodic_strips_non_uniform.hpp`  is added. 

---

### All Submissions

- [ ] Do you follow the conventions specified in our [coding standards](https://gyselax.github.io/gyselalibxx/docs/standards/CODING_STANDARD.html) ?

### New Feature Submissions

### Changes to the CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add methods to `PolarSplineEvaluator` to avoid unnecessary creation of fields of coordinates.
 - Allow a `DerivField` to be stored in a `MultipatchField`.
 - Add `DerivFieldOnPatch` and `IdxRangeSliceOnPatch` aliases for the `MultipatchField` in `types.hpp`.
-- Add a `periodic_strips_non_uniform_2d_9patches` geometry. 
-
+- Add a `periodic_strips_non_uniform_2d_9patches` geometry.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Port `PolarSplineEvaluator` methods to GPU.
 - Add methods to `PolarSplineEvaluator` to avoid unnecessary creation of fields of coordinates.
 - Allow a `DerivField` to be stored in a `MultipatchField`.
-- Add `DerivFieldOnPatch` and `IdxRangeSliceOnPatch` aliases for the `MultipatchField` in `types.hpp`.
 - Add a `periodic_strips_non_uniform_2d_9patches` geometry.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clean up code in `BslExplicitPredCorrRTheta::operator()`.
 - Clean up code in `BslImplicitPredCorrRTheta::operator()`.
 - Remove the const version of `get_values_field()` and change it into `get_values_const_field()` in `DerivFieldCommon`.
-- 
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Port `PolarSplineEvaluator` methods to GPU.
 - Add methods to `PolarSplineEvaluator` to avoid unnecessary creation of fields of coordinates.
+- Add `DerivFieldOnPatch` aliases for the `MultipatchField` in `types.hpp`. 
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add methods to `PolarSplineEvaluator` to avoid unnecessary creation of fields of coordinates.
 - Allow a `DerivField` to be stored in a `MultipatchField`.
 - Add `DerivFieldOnPatch` and `IdxRangeSliceOnPatch` aliases for the `MultipatchField` in `types.hpp`.
+- Add a `periodic_strips_non_uniform_2d_9patches` geometry. 
 
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


PaulineVidal:
@EmilyBourne For the tests for `InterfaceDerivativeMatrix`, I prefer to use non-uniform grids to better identify potential bugs or indexation errors.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add methods to `PolarSplineEvaluator` to avoid unnecessary creation of fields of coordinates.
 - Allow a `DerivField` to be stored in a `MultipatchField`.
 - Add `DerivFieldOnPatch` and `IdxRangeSliceOnPatch` aliases for the `MultipatchField` in `types.hpp`.
+- Add a `periodic_strips_non_uniform_2d_9patches` geometry.
```
EmilyBourne, line 18: Is this a new geometry or a new test? Remember the changelog is here to explain to users what they need to change to update their code or what new features they can now use in their code

PaulineVidal: A new geometry. It is not supposed to change the previous test.

EmilyBourne (commented):
The code looks fine to me. But I am still not clear on what can be tested with this geometry that couldn't be tested with any of the existing geometries. The multiplication of these geometries is not good for maintainability and in particular the cases with a large number of patches are pain points as the tests are very verbose (in order to initialise everything).

It seems to me that cases requiring a large number of patches are tested by the existing `9patches_2d_periodic_strips_uniform.hpp` and cases with non-uniform points are tested by the existing `3patches_2d_non_periodic_non_uniform.hpp`, `2patches_2d_onion_shape_non_uniform.hpp` or `2patches_2d_non_periodic_non_uniform.hpp`. What am I missing? What tests require this geometry?

Or inversely if this is required what tests require `9patches_2d_periodic_strips_uniform.hpp` but couldn't be tested with the existing uniform cases plus a non-uniform version of this test?

CHANGELOG.md
EmilyBourne: Geometries that affect users should have their own sub-folder in the `src/` folder. I don't think this is a new geometry from a user view point.

PaulineVidal:
It is better to test the `InterfaceDerivativeMatrix` on 9 patches (to check all the first derivatives on x and on y and the cross-derivatives are well computed, so to have at least 3 patches in each direction). The uniform case can hide indexation error. For example, for the following geometry, 
```
| 1 | 2 | 3 |
``` 
if the patches are all uniform with the same number of points, the computation of the coefficients at the interface 1|2 and the ones at the inteface 2|3 will give the same values. So, it can hide errors in the code.

PaulineVidal:
> > It is better to test the `InterfaceDerivativeMatrix` on 9 patches (to check all the first derivatives on x and on y and the cross-derivatives are well computed, so to have at least 3 patches in each direction). The uniform case can hide indexation error. For example, for the following geometry,
> > ```
> > | 1 | 2 | 3 |
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > if the patches are all uniform with the same number of points, the computation of the coefficients at the interface 1|2 and the ones at the inteface 2|3 will give the same values. So, it can hide errors in the code.
> > And here, it doesnt add a lot of information from the previous geometry as everything depends on `9patches_2d_periodic_strips.hpp`
> 
> I'm not sure this answers my question. From your description I get the impression that if there were 2 tests instead of 1 then everything could be tested with existing geometries? This could also help isolate which problems are indexation errors and which are related to directions

With only one direction, it wont test the computation of the cross-derivatives.

EmilyBourne:
> > > It is better to test the `InterfaceDerivativeMatrix` on 9 patches (to check all the first derivatives on x and on y and the cross-derivatives are well computed, so to have at least 3 patches in each direction). The uniform case can hide indexation error. For example, for the following geometry,
> > > ```
> > > | 1 | 2 | 3 |
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > if the patches are all uniform with the same number of points, the computation of the coefficients at the interface 1|2 and the ones at the inteface 2|3 will give the same values. So, it can hide errors in the code.
> > > And here, it doesnt add a lot of information from the previous geometry as everything depends on `9patches_2d_periodic_strips.hpp`
> > 
> > 
> > I'm not sure this answers my question. From your description I get the impression that if there were 2 tests instead of 1 then everything could be tested with existing geometries? This could also help isolate which problems are indexation errors and which are related to directions
> 
> With only one direction patch splitting, it wont test the computation of the cross-derivatives.

Right. So you would have 1 test to test indexation errors using `3patches_2d_non_periodic_non_uniform.hpp` and another to test cross-derivatives using `9patches_2d_periodic_strips_uniform.hpp`?

PaulineVidal:
The interesting thing with the 9 patches is that the 
```
| 1 | 2 | 3 |
| 4 | 5 | 6 |
| 7 | 8 | 9 |
```
all the information of the patch5 are computed with the method and do not rely on the boundary conditions.

EmilyBourne:
> The interesting thing with the 9 patches is that the
> 
> ```
> | 1 | 2 | 3 |
> | 4 | 5 | 6 |
> | 7 | 8 | 9 |
> ```
> 
> all the information of the patch5 are computed with the method and do not rely on the boundary conditions.

Ok. And there are differences for the non-uniform case that are relevant to this?
Maybe this would bother me less if we could remove one or two of the existing cases (e.g. 2 patch cases?) or if we could avoid the need to copy-paste the initialisation of these big cases every time

PaulineVidal:
> > The interesting thing with the 9 patches is that the
> > ```
> > | 1 | 2 | 3 |
> > | 4 | 5 | 6 |
> > | 7 | 8 | 9 |
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > all the information of the patch5 are computed with the method and do not rely on the boundary conditions.
> 
> Ok. And there are differences for the non-uniform case that are relevant to this? Maybe this would bother me less if we could remove one or two of the existing cases (e.g. 2 patch cases?) or if we could avoid the need to copy-paste the initialisation of these big cases every time

The non-uniform case allows us to check that we are not mixing the coefficient with other patches. 
Why do you mean about removing the 2 patch cases? The other geometries are relying on the two patch geometry. 
image

The fact is also that the physical case will be complicated with a lot of patches. But maybe we can implement functions like in `input` that initialise everything?

EmilyBourne (approved):

PR #414, opened 2025-08-26
Add curl operator

yasahi-hpc:
Add a curl operator working on arbitrary geometry.

---

### All Submissions

- [ ] Have you linked any issues that should be closed when this PR is merged (using closing keywords) ?

### New Feature Submissions

- [x] Have you added tests for the new functionalities ?

### Changes to the CI

### DIFF
--- a/src/math_tools/curl.hpp
+++ b/src/math_tools/curl.hpp
@@ -8,7 +8,6 @@
 #include "static_tensors.hpp"
 #include "vector_field.hpp"
 
-
 /**
  * @brief A class which implements a curl operator
  * The implemented equation is:
@@ -35,24 +34,18 @@ class Curl
 private:
     Mapping3D m_mapping;
     MetricTensorEvaluator<Mapping3D, MappingCoord> m_metric_tensor;
-    Gradient<MetricTensorEvaluator<Mapping3D, MappingCoord>> m_grad;
 
 public:
     /**
-     * @brief Build a LiePoissonBracket operator.
+     * @brief Build a curl operator.
      * @param mapping The mapping describing the system of coordinates on which the
      *              expression is calculated.
      */
-    explicit Curl(Mapping3D const& mapping)
-        : m_mapping(mapping)
-        , m_metric_tensor(mapping)
-        , m_grad(m_metric_tensor)
-    {
-    }
+    explicit Curl(Mapping3D const& mapping) : m_mapping(mapping), m_metric_tensor(mapping) {}
 
     /**
-     * @brief Compute the gyrokinetic Poisson bracket at a given coordinate, from
-     * the partial derivatives of the two fields, and the magnetic field.
+     * @brief Compute a curl at a given coordinate, from
+     * the partial derivatives of the vector field f
      *
      * @param[in] partial_derivatives_f A tensor containing the partial derivatives
      * of the vector field f expressed at the given coordinate.
--- a/src/math_tools/curl.hpp
+++ b/src/math_tools/curl.hpp
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include "ddc_alias_inline_functions.hpp"
+#include "ddc_aliases.hpp"
+#include "gradient.hpp"
+#include "metric_tensor_evaluator.hpp"
+#include "static_tensors.hpp"
+#include "vector_field.hpp"
+
+
+/**
+ * @brief A class which implements a curl operator
+ * The implemented equation is:
+ * @f$ \nabla \cross \mathbf{F} @f$
+ * @f$ \nabla \cross \mathbf{F} = {\cal J}_{\rm x}^{-1}\epsilon^{klm}\partial_{x^l} F_m @f$
+ * with @f$ {\cal J}_{\rm x} @f$ the jacobian of the system,
+ * @f$ F_m @f$ the covariant components of F and @f$\epsilon^{klm} @f$ the Levi-Civita symbol.
+ * @tparam Mapping3D A type representing a mapping in 3 dimensions.
+ * @tparam MappingCoord The type of the coordinate that will be used to evaluate the
+ *                  mapping. This coordinate is used to calculate the determinant of
+ *                  the Jacobian and the metric tensor. It is almost always equal to
+ *                  the argument type of the 3D mapping but axi-symmetry may make it
+ *                  useful to evaluate the mapping on a coordinate with a reduced
+ *                  dimensionality.
+ */
+template <class Mapping3D, class MappingCoord = typename Mapping3D::CoordArg>
+class Curl
+{
+    static_assert(is_mapping_v<Mapping3D>);
+
+    using BasisSpatial = ddc::to_type_seq_t<typename Mapping3D::CoordArg>;
+    using CovBasisSpatial = get_covariant_dims_t<BasisSpatial>;
+
+private:
+    Mapping3D m_mapping;
+    MetricTensorEvaluator<Mapping3D, MappingCoord> m_metric_tensor;
+    Gradient<MetricTensorEvaluator<Mapping3D, MappingCoord>> m_grad;
+
+public:
+    /**
+     * @brief Build a LiePoissonBracket operator.
+     * @param mapping The mapping describing the system of coordinates on which the
+     *              expression is calculated.
+     */
+    explicit Curl(Mapping3D const& mapping)
+        : m_mapping(mapping)
+        , m_metric_tensor(mapping)
+        , m_grad(m_metric_tensor)
+    {
+    }
+
+    /**
+     * @brief Compute the gyrokinetic Poisson bracket at a given coordinate, from
+     * the partial derivatives of the two fields, and the magnetic field.
+     *
+     * @param[in] partial_derivatives_f A tensor containing the partial derivatives
+     * of the vector field f expressed at the given coordinate.
+     * @param[in] coord The coordinate where the calculation is carried out.
+     */
+    template <
+            class TensorType,
+            class = std::enable_if_t<is_tensor_type_v<TensorType> && TensorType::rank() == 2>>
+    KOKKOS_INLINE_FUNCTION auto operator()(
+            TensorType const& partial_derivatives_f,
+            MappingCoord const& coord) const
+    {
+        double J = m_mapping.jacobian(coord);
+        LeviCivitaTensor<double, BasisSpatial> eps(J);
+        return tensor_mul(index<'k', 'l', 'm'>(eps), index<'l', 'm'>(partial_derivatives_f));
+    }
+};
--- a/src/math_tools/curl.hpp
+++ b/src/math_tools/curl.hpp
@@ -53,7 +53,7 @@ public:
      */
     template <class T, class CovBasisF>
     KOKKOS_INLINE_FUNCTION Tensor<T, get_contravariant_dims_t<CovBasisF>> operator()(
-            Tensor<T, CovBasisSpatial, CovBasisF> const& partial_derivatives_f,
+            Tensor<T, CovBasisF, CovBasisSpatial> const& partial_derivatives_f,
             MappingCoord const& coord) const
     {
         static_assert(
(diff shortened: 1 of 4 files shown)


src/math_tools/curl.hpp
```diff
@@ -0,0 +1,65 @@
...
+            class TensorType,
+            class = std::enable_if_t<is_tensor_type_v<TensorType> && TensorType::rank() == 2>>
+    KOKKOS_INLINE_FUNCTION auto operator()(
+            TensorType const& partial_derivatives_f,
+            MappingCoord const& coord) const
+    {
+        double J = m_mapping.jacobian(coord);
+        LeviCivitaTensor<double, BasisSpatial> eps(J);
+        return tensor_mul(index<'k', 'l', 'm'>(eps), index<'l', 'm'>(partial_derivatives_f));
+    }
```
EmilyBourne, line 64: Is it possible to be more specific about the types here? I am concerned that the errors may be a little opaque. E.g. if a user passes a contravariant object as `partial_derivatives_f` instead of a covariant object they will get an error such as:
```
Index is not compatible with tensor type
```
coming from code that they didn't write

src/math_tools/curl.hpp
```diff
@@ -0,0 +1,65 @@
...
+    template <
+            class TensorType,
+            class = std::enable_if_t<is_tensor_type_v<TensorType> && TensorType::rank() == 2>>
+    KOKKOS_INLINE_FUNCTION auto operator()(
+            TensorType const& partial_derivatives_f,
+            MappingCoord const& coord) const
+    {
+        double J = m_mapping.jacobian(coord);
+        LeviCivitaTensor<double, BasisSpatial> eps(J);
+        return tensor_mul(index<'k', 'l', 'm'>(eps), index<'l', 'm'>(partial_derivatives_f));
```
EmilyBourne, line 63: According to the [Coding Standards](https://gyselax.github.io/gyselalibxx/docs/standards/CODING_STANDARD.html)
> The auto keyword should be avoided where possible. It is permitted in one of the following cases:
> - to store objects that can not be typed otherwise (e.g. lambdas)
> - to store the result of an expression specifically specifying the type of the generated value

tests/math_tools/test_curl.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    double eps = 1e-12;
+    EXPECT_NEAR(ddcHelper::get<Rho>(curl_B), 0.0, eps);
+    EXPECT_NEAR(ddcHelper::get<Theta>(curl_B), 0.0, eps);
+    EXPECT_NEAR(ddcHelper::get<Phi>(curl_B), 0.0, eps);
+}
+
+// Test the curl for a simple B-field where the
+// result can be analytically computed.
+// Let's use B = (0, 0, 1/(R0+r))
```
EmilyBourne, line 75: Is this expression in contravariant components?

tests/math_tools/test_curl.cpp
```diff
@@ -0,0 +1,134 @@
...
+    using CovBasisSpatial = get_covariant_dims_t<BasisSpatial>;
+    IdxStepRho nrho(10);
+    IdxStepTheta ntheta(10);
+    IdxStepPhi nphi(2);
+    ddc::init_discrete_space<GridRho>(
+            GridRho::init<GridRho>(Coord<Rho>(0.0), Coord<Rho>(1.0), nrho));
+    ddc::init_discrete_space<GridTheta>(GridTheta::init<GridTheta>(
+            build_uniform_break_points(Coord<Theta>(0.0), Coord<Theta>(2 * M_PI), ntheta)));
+    ddc::init_discrete_space<GridPhi>(
+            GridPhi::init<GridPhi>(Coord<Phi>(0.0), Coord<Phi>(2 * M_PI), nphi));
```
EmilyBourne, line 101: Are these grids used?

tests/math_tools/test_curl.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    Coord<Rho, Theta, Phi> coord
+            = ddc::coordinate(Idx<GridRho, GridTheta, GridPhi>(nrho - 1, ntheta - 5, 0));
```
EmilyBourne, line 49: Are the calls to `init_discrete_space` necessary? Couldn't you simply define a test coordinate?

EmilyBourne (changes requested):
Don't forget to check through the checklist in the PR description.
Please remove the draft status when the PR is ready for review

src/math_tools/curl.hpp
yasahi-hpc: I am also considering that.
If I can extract the underlying types of Tensor, I can add a static_assert to check that the tensor is <cov, cov>.

src/math_tools/curl.hpp
yasahi-hpc: I just copy pasted from https://github.com/gyselax/gyselalibxx/blob/9a4e83ae67b2b99c8e1582e48cb88e486aa2d240/src/math_tools/lie_poisson_bracket.hpp#L65-#L73

I thought there is a specific reason to use auto. If not, I can hardcode the type

tests/math_tools/test_curl.cpp
yasahi-hpc: I actually do not understand how it works. I only need a valid coordinate (point).

src/math_tools/curl.hpp
EmilyBourne:
```suggestion
    template<class T, class CovBasisF>
    KOKKOS_INLINE_FUNCTION Tensor<T, get_contravariant_dims_t<CovBasisF>> operator()(
            Tensor<T, CovBasisF, CovBasisSpatial> const& partial_derivatives_f,
            MappingCoord const& coord) const
    {
        static_assert(is_covariant_vector_index_set<CovBasisF>, "Expected the derivative of f expressed in the covariant basis");
        double J = m_mapping.jacobian(coord);
        LeviCivitaTensor<double, BasisSpatial> eps(J);
        return tensor_mul(index<'k', 'l', 'm'>(eps), index<'l', 'm'>(partial_derivatives_f));
    }
```
?

src/math_tools/curl.hpp
EmilyBourne: Oops, probably I used auto in my first implementation and forgot to go back. This should have been picked up in review :|

src/math_tools/curl.hpp
```diff
@@ -0,0 +1,66 @@
...
+     * @brief Compute a curl at a given coordinate, from
+     * the partial derivatives of the vector field f
+     *
+     * @param[in] partial_derivatives_f A tensor containing the partial derivatives
+     * of the vector field f expressed at the given coordinate.
+     * @param[in] coord The coordinate where the calculation is carried out.
+     */
+    template <class T, class CovBasisF>
+    KOKKOS_INLINE_FUNCTION Tensor<T, get_contravariant_dims_t<CovBasisF>> operator()(
+            Tensor<T, CovBasisSpatial, CovBasisF> const& partial_derivatives_f,
```
EmilyBourne, line 56: `partial_derivatives_f` is similar to a Jacobian matrix right?
image

In which case this seems like the transpose of what we want. Usually the element $(i,j)$ of a Jacobian matrix is defined as $\frac{\partial f_i}{\partial x_j}$. Here you have: $\frac{\partial f^j}{\partial x_i}$

src/math_tools/curl.hpp
EmilyBourne: > since B may be cov or contra.

Ok. In that case this type shouldn't have `Cov` in the name

src/math_tools/curl.hpp
```diff
@@ -0,0 +1,66 @@
...
+            MappingCoord const& coord) const
+    {
+        static_assert(
+                is_covariant_vector_index_set_v<CovBasisF>,
+                "Expected the derivative of f expressed in the covariant basis");
```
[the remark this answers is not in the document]
yasahi-hpc: I would say Expected the derivative of: f expressed in the covariant basis.
I mean \grad_{l} F_{m}. 

> If you mean the second one then doesn't this contradict what you said earlier?
> 
> B may be cov or contra

What I mean is that we can only accept \grad_{l} B_{m}. People are more likely to make a mistake on B, like \grad_{l} B^{m} rather than \grad^{l} B_{m}. I am thus inclined to error out \grad_{l} B^{m}.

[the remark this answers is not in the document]
EmilyBourne: So: $\nabla F_m$

But I don't think you handle both $\nabla^l F_m$ and $\partial_l F_m$ do you? You would need an additional multiplication by the metric tensor for the former case.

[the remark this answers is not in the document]
yasahi-hpc: I can only manipulate $\partial_l F_m$.
I just compute $\left(\nabla \times F\right)^k = \frac{1}{\sqrt{g}}\epsilon^{klm} \nabla_{l} F_{m}$

src/math_tools/curl.hpp
EmilyBourne: If you want your class to work on the cartesian basis you may need to do the same

yasahi-hpc: That makes sense. In my tests, I do not use the identity tensor

EmilyBourne: Actually it is not even only in the tests. There are multiple places in Gysela where we have $\\{F, \hat{e}_i\\}$ where $\hat{e}_i$ are the unit vectors so I needed this use case for Lie-Poisson

tests/math_tools/test_curl.cpp
```diff
@@ -0,0 +1,116 @@
...
+    double const theta = ddc::get<Theta>(coord);
+    double const cos_t = std::cos(theta);
+    double const sin_t = std::sin(theta);
+
+    // Compute dB_phi/dr, dB_phi/dtheta
+    // B_i needs to be computed by g_ij B^j
+    ddcHelper::get<Rho_cov, Rho_cov>(grad_B) = 0.0;
+    ddcHelper::get<Rho_cov, Theta_cov>(grad_B) = 0.0;
+    ddcHelper::get<Rho_cov, Phi_cov>(grad_B)
+            = (R0 + r * cos_t) * (-R0 - r * cos_t + 2 * (R0 + r) * cos_t) / ((R0 + r) * (R0 + r));
```
EmilyBourne, line 100: Following usual conventions for the Jacobian matrix these 3 values should represent the 3 derivatives $\partial_r B_r$, $\partial_\theta B_r$, $\partial_\phi B_r$

This doesn't seem to match your definition though

EmilyBourne: Ok. Then let's use the existing convention :+1: you will need to swap the indices in the curl operator

yasahi-hpc: OK. So, what does this do mathematically?

```C++
tensor_mul(index<'k', 'l', 'm'>(eps), index<'l', 'm'>(partial_derivatives_f));
```
I thought this is $\left(\nabla \times F\right)^k = \frac{1}{\sqrt{g}}\epsilon^{klm} \nabla_{l} F_{m}$

src/math_tools/curl.hpp
```diff
@@ -0,0 +1,68 @@
...
+        double J = m_mapping.jacobian(coord);
+        LeviCivitaTensor<double, BasisSpatial> eps(J);
+
+        // The following code computes: 1/sqrt(g) * eps^{klm} \nabla_{l} F_{m}
+        return tensor_mul(index<'k', 'l', 'm'>(eps), index<'m', 'l'>(partial_derivatives_f));
```
EmilyBourne, line 66: The comment in the class description needs updating

src/math_tools/curl.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include "gradient.hpp"
+#include "metric_tensor_evaluator.hpp"
+#include "static_tensors.hpp"
+#include "vector_field.hpp"
+
+/**
+ * @brief A class which implements a curl operator
+ * The implemented equation is:
+ * @f$ \nabla \cross \mathbf{F} @f$
+ * @f$ \nabla \cross \mathbf{F} = {\cal J}_{\rm x}^{-1}\epsilon^{klm}\partial_{x^l} F_m @f$
```
EmilyBourne, line 15:
```suggestion
 * @f$ \nabla \cross \mathbf{F} = {\cal J}_{\rm x}^{-1}\epsilon^{klm}\partial_{x^m} F_l @f$
```

EmilyBourne: Never mind sorry. I've just checked this this morning and I am wrong. The docstring is fine.

PR #466, opened 2025-11-25
Update Spack environments

tpadioleau:
This PR updates Spack in version 1 in all Spack-based toolchains.

- Increase the `DISCOVERY_TIMEOUT` to avoid getting timeouts on supercomputers.

# All toolchains

- Use the new `repos` section to automatically clone extra repositories
- Use v2025.11 packages by default
- Improve uniformity of the specs between toolchains
- Set `PYTHONPYCACHEPREFIX` early
- Updates:
  - Kokkos 4.7
  - Python 3.12:

# Jean-Zay toolchain

- Patch `arrow` to always disable `ARROW_MIMALLOC` (otherwise needs internet access on a compute node during cmake configuration stage)
- Use the new `--concurrent-packages`

# Adastra toolchain

- Use modules `develop` and `spack-user-5.0.0`
- Patch `pdiplugin-netcdf` to add the MPI dependency.
- Use raw `gcc` (no CRAY wrapper) but needs a dirty fix in the `toolchain.cmake` file to handle GPU direct
- Use `openblas` instread of `cray-libsci`
- Use the new `--concurrent-packages` instead of manually spawning `spack install` commands

# CPU dev toolchain

- Disable `python` variant of `pdi` on macOS
- Disable `performance` variant of `py-pandas` on Ubuntu to avoid a dependency to `llvm`. It concretized with `llvm@16` but it is not reported as support in the official documentation, see https://numba.readthedocs.io/en/stable/user/installing.html#version-support-information.

---

### All Submissions

### New Feature Submissions

### Changes to the CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@ The operator averages the values of the advection field on the first ring to get
 - Make the choice of equililibrium an input parameter for `vlasovpoisson_xvx` executables.
 - Move static class method `BumpontailEquilibrium::init_from_input` to namespace `bumpontail_equilibrium::init_from_input`.
 - Move static class method `MaxwellianEquilibrium::init_from_input` to namespace `maxwellian_equilibrium::init_from_input`.
+- Update CPU Spack toolchain to Spack v1 and update packages, for example Kokkos 4.7, Python 3.12:.
+- Update Adastra Spack toolchains to Spack v1 and update packages, for example Kokkos 4.7, Python 3.12:. Use OpenBLAS instead of Cray LibSci and raw GCC compilers.
+- Update Jean-Zay Spack toolchain to Spack v1 and update packages, for example Kokkos 4.7, Python 3.12:.
 
 ### Deprecated
 


toolchains/cpu.spack.gyselalibxx_env/gyselalibxx-env-1.1.0.yaml
```diff
@@ -26,18 +26,32 @@ spack:
...
+    blas:
+      require: openblas@0.3
+    lapack:
+      require: openblas@0.3
+    mpi:
+      require: openmpi@5.0
+    pdi:
+      require: ~python
+    py-pandas:
+      require: ~performance
```
EmilyBourne: What decides what is specified here rather than above? The positive requirements (e.g. `+mpi`) are specified above but the negative requirements are here. Is there a more fundamental difference between these kinds of requirements?

EmilyBourne:
> The failure on the Ubuntu installation is only due to a slightly larger error than the threshold. IMO we can safely ignore.

1e-14 seems low as a tolerance on a derivative for splines. I think we could safely increase to 1e-13

EmilyBourne:
> > > The failure on the Ubuntu installation is only due to a slightly larger error than the threshold. IMO we can safely ignore.
> > 
> > 
> > 1e-14 seems low as a tolerance on a derivative for splines. I think we could safely increase to 1e-13
> 
> Should I do it in this PR ?

Yes I think we can. Otherwise this will just pop up again in later PRs

PR #488, opened 2025-12-16
Use C++ 20 concepts to simplify coordinate transformation description

EmilyBourne:
We need to move to a more modern version of C++ to prepare for Kokkos v5. Nothing in our workflow currently prevents this. This PR therefore increases the C++ version to C++20. This allows us to use concepts. Concepts are used to simplify the interface tests for coordinate transformations. The resulting code is easier to read, and better enforces the interface. As a result 2 missing `const` annotations are added.

The naming conventions chosen for concepts are:
- Create a `concepts` namespace to store all concepts used to describe types (e.g. `template<concepts::Mapping Mapping>`
- Use aliases to create concepts with clearer names for static assertions (e.g. `template<class T> concept is_mapping_v = concepts::Mapping<T>;`)

---

### All Submissions

### New Feature Submissions

### Changes to the CI

### DIFF
--- a/src/coord_transformations/combined_mapping.hpp
+++ b/src/coord_transformations/combined_mapping.hpp
@@ -94,7 +94,7 @@ public:
      *
      * @return The equivalent coordinate.
      */
-    CoordResult operator()(CoordArg coord)
+    CoordResult operator()(CoordArg coord) const
     {
         return m_mapping_1(m_mapping_2(coord));
     }


EmilyBourne:
@tpadioleau any idea what is going wrong on GPU? It seems that the compiler has a bug. Is this known? Do you know if concepts are supposed to work on GPU?

tpadioleau:
Can you have a look in verbose mode ? It is not impossible that kokkos imposes the standard to downstream projects and kokkos was compiled in C++17 mode.

tpadioleau:
If we really want it, we may need to update Persee to Kokkos 4.7.02.

EmilyBourne:
> If we really want it, we may need to update Persee to Kokkos 4.7.02.

GMGPolar is pinned to C++20. I planned to use concepts to remove the inheritance (ready for GPU). It would be useful to at least confirm if concepts ought to work so I don't write code that won't be valid.

tpadioleau:
> > If we really want it, we may need to update Persee to Kokkos 4.7.02.
> 
> GMGPolar is pinned to C++20. I planned to use concepts to remove the inheritance (ready for GPU). It would be useful to at least confirm if concepts ought to work so I don't write code that won't be valid.

I think we should be fine, Nvidia started the support of C++20 in CUDA 12, it was a mess at first but now should be fine.

PR #596, opened 2026-03-18
v0.7.0

EmilyBourne:


tpadioleau (approved):

tpadioleau:
> > I am not sure to get why we have these PR actually :) Is it explained somewhere ?
> 
> It is not explained. The only reason is to avoid giving people the rights to push to main

Ok, I would be interested in discussing release workflow at some point.

PR #666, opened 2026-05-13
Allow Landau4D to run with Lagrange interpolation

EmilyBourne:
Allow Landau4D to run with Lagrange interpolation.

**Commit Summary**
- Add `lagrange_definitions_xyvxvy.hpp` to provide Lagrange definitions
- Add an extra compilation target for `landau4d_fft.cpp`
- Change quadrature for Landau4D to Simpson quadrature (with Neumann quadrature at the start of the domain if the number of cells is not compatible). This ensures splines are not needed
- Use macros to define the interpolation and initialise the grids in Landau4D to differentiate between Spline and Lagrange
- Generate a different default yaml file when using Lagrange interpolation
- Add a missing type declaration `DFieldMemVy`
- Add an optional `IdxRange` argument to the constructor of `LagrangeInterpolator` to make it interchangeable with `SplineInterpolator`
- Add a new method `init_lagrange_dependent_idx_range`
- Test execution of the Lagrange version of Landau4D

---

### All Submissions

### New Feature Submissions

### Changes to the CI


simulations/geometryXYVxVy/landau/params.yaml.hpp
```diff
@@ -32,3 +33,38 @@ constexpr char const* const params_yaml = R"PDI_CFG(SplineMesh:
 Output:
   time_diag: 0.24
 )PDI_CFG";
+
```
gdgirard: There is no way to avoid to duplicate all these lines for a change that is limitated to the number of cells in vpar direction.

gdgirard (approved):

simulations/geometryXYVxVy/landau/params.yaml.hpp
EmilyBourne: It is not trivial as macros cannot be placed inside strings. It is possible with the following code:
```
constexpr char const* const params_yaml =
#ifdef SPLINE
        "SplineMesh:"
#elif defined(LAGRANGE)
        "Mesh:"
#endif
        R"PDI_CFG(
  x_min: 0.0
  x_max: 12.56637061435917
  x_ncells: 32
  y_min: 0.0
  y_max: 12.56637061435917
  y_ncells : 32
  vx_min: -6.0
  vx_max: +6.0
)PDI_CFG"
#ifdef SPLINE
        "  vx_ncells: 33"
#elif defined(LAGRANGE)
        "  vx_ncells: 31"
#endif
        R"PDI_CFG(
  vy_min: -6.0
  vy_max: +6.0
)PDI_CFG"
#ifdef SPLINE
        "  vy_ncells: 33"
#elif defined(LAGRANGE)
        "  vy_ncells: 31"
#endif
        R"PDI_CFG(

SpeciesInfo:
- charge: -1.
  mass: 0.0005
  density_eq: 1.
  temperature_eq: 1.
  mean_velocity_eq: 0.
  perturb_amplitude: 0.05
  perturb_mode: 1

Algorithm:
  deltat: 0.12
  nbiter: 140

Output:
  time_diag: 0.24
)PDI_CFG";
```
However I am not convinced that this is better as it is much harder to read. Further it is quite difficult to understand what belongs to which indentation level. Which version do you prefer?

gdgirard: Yes I agree with you. Is it explain somewhere why you need 33 cells in the case of splines and 31 in the case of Lagrange ? I think it would be good to add this kind of information as a comment in this default YAML file. Idem for the other dimension if you explicitely need a power of 2 for periodic directions.

EmilyBourne: I'm not sure that this is compulsory. I wanted to have a power of 2 by default for the Lagrange test to make it easier to test with MPI. I am not sure why there are 33 cells in the spline case. This seems like a strange choice to me

gdgirard: If 33 is not mandatory and 31 is allowed it would simplify the YAML file and avoid the IFDEF and this copy-paste ?

EmilyBourne: It seems to run. I have changed this: [591c86d](https://github.com/gyselax/gyselalibxx/pull/666/commits/591c86d25e0a68dd3c437024db4ec544992e2e7d)

gdgirard (approved):

PR #695, opened 2026-06-29
Allow polar advection to modify a subset of the distribution function

EmilyBourne:
Allow polar advection to modify a subset of the distribution function (e.g. to fix outer bounds)

---

### All Submissions

### New Feature Submissions

### Changes to the CI

### DIFF
--- a/src/advection/bsl_advection_polar.hpp
+++ b/src/advection/bsl_advection_polar.hpp
@@ -120,8 +120,6 @@ private:
 
     LogicalToPhysicalMapping const& m_logical_to_physical_mapping;
 
-    std::optional<IdxRangeBatched> m_idx_range_advected_points;
-
 public:
     /**
      * @brief Instantiate an advection operator.
@@ -142,13 +140,11 @@ public:
             Builder2D const& builder_2d,
             Evaluator2D const& evaluator_2d,
             FootFinder& foot_finder,
-            LogicalToPhysicalMapping const& logical_to_physical_mapping,
-            std::optional<IdxRangeBatched> idx_range_advected_points = std::nullopt)
+            LogicalToPhysicalMapping const& logical_to_physical_mapping)
         : m_builder_2d(builder_2d)
         , m_evaluator_2d(evaluator_2d)
         , m_find_feet_method(foot_finder)
         , m_logical_to_physical_mapping(logical_to_physical_mapping)
-        , m_idx_range_advected_points(idx_range_advected_points)
     {
     }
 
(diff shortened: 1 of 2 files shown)


gdgirard (approved):

src/advection/bsl_advection_polar.hpp
```diff
@@ -163,13 +162,17 @@ class BslAdvectionPolar
      *      of the advection field at each point on the logical grid.
      * @param [in] dt
      *      A time step used.
+     * @param[in] idx_range_advected_points
+     *      The index range of the parts of the distribution function which are
+     *      modified. This allows regions (e.g. boundary regions) to remain constant.
```
EmilyBourne: Should this be an argument to the `operator()`, passed at each call, or a constructor argument?

The default index range comes from `allfdistribu` so keeping it in `operator()` theoretically allows the same operator to be used for different batches. In practice though the FootFinder defines the batch dims anyway and putting it in the constructor ensures that all calls are using the same method. This may be safer?

tpadioleau (commented):
It does not work if we do a slice on the caller site ?

Not blocking

EmilyBourne:
> It does not work if we do a slice on the caller site ?
> 
> Not blocking

I don't think so as this would mess with the internal call to the spline builder. The values that aren't modified are still used to construct the spline representation

src/advection/bsl_advection_polar.hpp
EmilyBourne: @gdgirard ?

tpadioleau:
> > It does not work if we do a slice on the caller site ?
> > Not blocking
> 
> I don't think so as this would mess with the internal call to the spline builder. The values that aren't modified are still used to construct the spline representation

I see, what about splitting the inout `DFieldFDistribu allfdistribu` parameter into two parameters in and out ? The out parameter could be smaller than the in parameter ?

EmilyBourne:
> > > It does not work if we do a slice on the caller site ?
> > > Not blocking
> > 
> > 
> > I don't think so as this would mess with the internal call to the spline builder. The values that aren't modified are still used to construct the spline representation
> 
> I see, what about splitting the inout `DFieldFDistribu allfdistribu` parameter into two parameters in and out ? The out parameter could be smaller than the in parameter ?

Wouldn't this be error-prone if you expect in to not be modified?

EmilyBourne:
> > > > It does not work if we do a slice on the caller site ?
> > > > Not blocking
> > > 
> > > 
> > > I don't think so as this would mess with the internal call to the spline builder. The values that aren't modified are still used to construct the spline representation
> > 
> > 
> > I see, what about splitting the inout `DFieldFDistribu allfdistribu` parameter into two parameters in and out ? The out parameter could be smaller than the in parameter ?
> 
> Wouldn't this be error-prone if you expect in to not be modified?

The restriction can be seen as imposing a boundary condition. Relevant discussion: https://github.com/gyselax/gyselalibxx/pull/695#discussion_r3490507784 ?

tpadioleau:
> Wouldn't this be error-prone if you expect in to not be modified?

Could be, not blocking anyway.

src/advection/bsl_advection_polar.hpp
gdgirard: This value won't change during the simulation, so it can be given to the constructor

PR #726, opened 2026-07-27
Add extrapolation to `GMGPolarPoissonLikeSolver` arguments

AbdelhadiKara:
Add extrapolation rule as a parameter of `GMGPolarPoissonLikeSolver` to set GMGPolar multigrid settings. By default the value is `Extrapolation::None`.
Add a private member m_extrapolation_rule to the `GMGPolarPoissonLikeSolver` class.

This fixes the extrapolation warning [seen in the poisson-mini-app](https://github.com/gyselax/gysela-mini-app_poisson/pull/7#issuecomment-4878593933).

---

### All Submissions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add a `extrapolation_rule` argument to `GMGPolarPoissonLikeSolver`. Default no extrapolation. 
+- Add a `extrapolation_rule` argument to `GMGPolarPoissonLikeSolver`. Default no extrapolation.
 - Add a new constructor for `GaussLegendre` from an index range describing the cell edges.
 - Add a `GradientCreator` operator to group derivative calculations.
 - Add a `NDLagrangeEvaluator` class.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add a `extrapolation_rule` argument to `GMGPolarPoissonLikeSolver`. Default no extrapolation.
+- Add an `extrapolation_rule` argument to `GMGPolarPoissonLikeSolver`. Default no extrapolation.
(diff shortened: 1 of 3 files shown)


EmilyBourne:
Please complete the checklist provided in the default PR description

julianlitz:
We could also implement a fallback to Extrapolation::None in GMGPolar when the mesh does not fulfill the Extrapolation criteria.

We already do this if no multigrid hierarchy can be achieved with the mesh.

Then you can just use extrapolation=1 for all cases.

EmilyBourne:
> We could also implement a fallback to Extrapolation::None in GMGPolar when the mesh does not fulfill the Extrapolation criteria.
> 
> We already do this if no multigrid hierarchy can be achieved with the mesh.
> 
> Then you can just use extrapolation=1 for all cases.

I would prefer to know which extrapolation is used. With the extrapolation as an argument we can put the choice in our input file which makes it easy for us to keep track of which method we used for which run

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add a `extrapolation_rule` argument to `GMGPolarPoissonLikeSolver`. Default no extrapolation.
```
EmilyBourne:
```suggestion
- Add an `extrapolation_rule` argument to `GMGPolarPoissonLikeSolver`. Default no extrapolation.
```

EmilyBourne (approved):

EmilyBourne:
We may need to tighten the tolerance to use the version without extrapolation and be sure that it gives sufficiently small errors
image

src/pde_solvers/gmg_polar_poisson_like_solver.hpp
```diff
@@ -181,6 +181,7 @@ class GMGPolarPoissonLikeSolver
     DomainGeometry const m_domain_geom;
     SplineBuilder const& m_builder;
     SplineEvaluator const& m_evaluator;
+    ExtrapolationType const& m_extrapolation_rule;
```
EmilyBourne: The constructor doesn't take a reference so this is a reference to the argument which goes out of scope at the end of the constructor.

PR #728, opened 2026-07-28
Make Gyselalib++ installable

### LINKED ISSUE #629: [FEATURE]: Make the library Gyselalib++ installable
In the current state of the library, if one runs `cmake --install build --prefix <prefix>` it only installs some of the executables of the directory `simulations` into `<prefix>/bin`. Even though it does not hurt, I think the main objective is to install the library part of the Gyselalib++ **library**.

It should allow users to consume Gyselalib++ in their CMake with something similar to:
```cmake
find_package(Gyselalibxx REQUIRED)
add_executable(my_fancy_simulation)
target_link_libraries(my_fancy_simulation PRIVATE gslx::geometry_xy gslx::vlasov)
```

tpadioleau:
This PR adds necessary commands to install Gyselalib++ as a CMake library:
- install header files with `target_sources`
- install CMake targets in the package `gyselalibxx`
- install a version config file with policy `SameMinorVersion`
- rename the CMake namespace `gslx` to `gyselalibxx` (breaking change)
- set C++20 in the interface of CMake targets

Regarding header files, their inclusion does not change yet. Headers with the same name, typically in geometry directories, are isolated into unique directories. From now on we also need to add header files to the CMakeLists.txt, otherwise the library will not be usable.

In practice this means users can do:
```cmake
find_package(gyselalibxx 0.7 REQUIRED)
target_link_libraries(some_executable PRIVATE gyselalibxx::advection)
```

---

### All Submissions

- [ ] Do you follow the conventions specified in our [coding standards](https://gyselax.github.io/gyselalibxx/docs/standards/CODING_STANDARD.html) ?

### New Feature Submissions

- [x] Have you added tests for the new functionalities ?

### Changes to the CI


src/geometryRTheta/advection_field/CMakeLists.txt
```diff
@@ -22,4 +22,4 @@ target_link_libraries("advection_field_RTheta" INTERFACE
 )
 add_library("gslx::advection_field_RTheta" ALIAS "advection_field_RTheta")
 
-install(TARGETS advection_field_RTheta EXPORT GyselalibxxTargets FILE_SET HEADERS)
+install(TARGETS advection_field_RTheta EXPORT GyselalibxxTargets FILE_SET HEADERS DESTINATION include/rtheta/advection_field)
```
EmilyBourne: I would be in favour of retaining the folder names. Alternatively we should rename the folders to match what you use here. Otherwise we are retaining all the VS Code link pitfalls

tpadioleau: I am not sure to understand what we mean by vscode pitfalls. In this version the user still needs to use `#include <iboltzmann.hpp>` and not `#include <xvx/boltzmann/boltzmann.hpp>`. The directories exist so that headers are isolated from each other, allowing the same name.

That said I agree that it may be more natural to keep the same names.

EmilyBourne: @gdgirard was suggesting that we move to using `#include <xvx/boltzmann/boltzmann.hpp>` as she (and sometimes vscode) struggle to work out how to find the imported file

EmilyBourne: Renaming would prevent us doing this later

tpadioleau: > Renaming would prevent us doing this later

I don't think we should rush into this. It will make the review more complex to read.

tpadioleau:
- We cannot have a mismatch between the project name and the namespace. Which one do we keep ?
  - `gyselalibxx` ?
  - `gslx` ?
- What should be the name of the package ? (the name we put in `find_package(<PackageName>)`)
  - `Gyselalibxx` ?
  - `gyselalibxx` ?
  - `gslx` ?

EmilyBourne:
> * We cannot have a mismatch between the project name and the namespace. Which one do we keep ?
>   
>   * `gyselalibxx` ?
>   * `gslx` ?
> * What should be the name of the package ? (the name we put in `find_package(<PackageName>)`)
>   
>   * `Gyselalibxx` ?
>   * `gyselalibxx` ?
>   * `gslx` ?

I would be in favour of `gyselalibxx` for both. I don't find `gslx` very explicit. It also looks a lot like an abbreviation of Gysela-X rather than Gyselalib++

src/geometryRTheta/advection_field/CMakeLists.txt
EmilyBourne: I am not clear on why having:
```
install(TARGETS advection_field_RTheta EXPORT GyselalibxxTargets FILE_SET HEADERS DESTINATION include/geometryRTheta/advection_field)
```
instead of
```
install(TARGETS advection_field_RTheta EXPORT GyselalibxxTargets FILE_SET HEADERS DESTINATION include/rtheta/advection_field)
```
would make the review more complex to read?

Unless you are referring to changing the includes which was never suggested for this PR?

You already seem to have made the change requested in this conversation though so I'm really not sure what your are worried about rushing into?

EmilyBourne:
Don't forget to complete the PR description and checklist before marking the PR as ready for review

tpadioleau:
> Don't forget to complete the PR description and checklist before marking the PR as ready for review

It is not ready yet, I just needed to remove the draft to start the CI

tpadioleau:
Testing it with Spack with custom branches:
```
[+] pigmql3 gyselalibxx@develop /local/home/tpadiole/spack-1.2.2/opt/spack/linux-x86_64/gyselalibxx-develop-pigmql3fmpxkhkoecp5jplpycfj3pku7 (2m19s)
[+] xhmtps4 gysela-xpp@develop /local/home/tpadiole/spack-1.2.2/opt/spack/linux-x86_64/gysela-xpp-develop-xhmtps4ughp3fgica64jfglt2yog6wc2 (1m52s)
```
(This ran `ctest` on Gysela-X++)

Testing it for the IO mini app, https://github.com/tpadioleau/dev-spack/actions/runs/30884222581/job/92619087815?pr=5 (Test ran `ctest` on Gyselalib++)

src/advection/CMakeLists.txt
```diff
@@ -2,18 +2,38 @@
 
 add_library("advection" INTERFACE)
 
-target_include_directories("advection"
-    INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}"
+target_compile_features("advection" INTERFACE cxx_std_20)
```
EmilyBourne, line 5: Does this now need adding to every library target? Is there no way to automatically have this across the project?

src/data_types/CMakeLists.txt
```diff
@@ -2,16 +2,35 @@
...
         DDC::core
         DDC::splines
-        gslx::utils
+        gyselalibxx::utils
 )
 
-add_library("gslx::data_types" ALIAS "data_types")
+add_library("gyselalibxx::data_types" ALIAS "data_types")
 
+install(TARGETS data_types EXPORT gyselalibxxTargets FILE_SET HEADERS)
```
EmilyBourne, line 36: Should the targets in `src/X` (with `X`!=`geometryY`) also specify the destination to preserve the entire folder structure?

src/geometryRTheta/initialisation/CMakeLists.txt
```diff
@@ -4,37 +4,49 @@ add_library(diocotron_init_eq STATIC
...
-        gslx::utils
+        DDC::pdi
+        gyselalibxx::geometry_RTheta
+        gyselalibxx::utils
 )
 
-add_library("gslx::diocotron_init_eq" ALIAS "diocotron_init_eq")
+add_library("gyselalibxx::diocotron_init_eq" ALIAS "diocotron_init_eq")
+
+install(TARGETS diocotron_init_eq EXPORT gyselalibxxTargets FILE_SET HEADERS)
```
EmilyBourne, line 26: Missing destination. I don't know if we need 2 libraries in this folder? I think we could combine the 2

src/advection/CMakeLists.txt
tpadioleau: Not that I know, that is what I remember a cmake presentation.

src/data_types/CMakeLists.txt
tpadioleau: I don't see the benefit apart from uniformity.

src/multipatch/connectivity/CMakeLists.txt
```diff
@@ -1,15 +1,29 @@
...
-    gslx::utils
+    gyselalibxx::coord_transformations
+    gyselalibxx::data_types
+    gyselalibxx::multipatch_data_types
+    gyselalibxx::utils
 )
-add_library("gslx::multipatch_connectivity" ALIAS "multipatch_connectivity")
+add_library("gyselalibxx::multipatch_connectivity" ALIAS "multipatch_connectivity")
+
+install(TARGETS multipatch_connectivity EXPORT gyselalibxxTargets FILE_SET HEADERS)
```
EmilyBourne, line 29: multipatch should definitely be treated like a kind of geometry

tpadioleau: Are there collisions with other targets ?

src/data_types/CMakeLists.txt
EmilyBourne: The benefit would be that we would be able to change to use `#include <data_types/File.hpp>` later if we want to

src/multipatch/connectivity/CMakeLists.txt
EmilyBourne: Probably

src/geometryRTheta/initialisation/CMakeLists.txt
tpadioleau: I am adding the destination, I let the question of combining them for later.

src/multipatch/connectivity/CMakeLists.txt
EmilyBourne: I can't see any that look dangerous currently apart from `src/multipatch/data_types/types.hpp` but I am not sure that this can be relied upon

src/advection/CMakeLists.txt
tpadioleau: I also wondered if we needed that many CMake targets ? They are currently not optional, the benefit is not clear.

src/multipatch/connectivity/CMakeLists.txt
EmilyBourne: But my point was mostly that conceptually it is very linked to geometries

src/advection/CMakeLists.txt
EmilyBourne: There could definitely be an argument for something like `gyselalibxx::core` combining all the folders in src/ except geometry* and multipatch

src/data_types/CMakeLists.txt
tpadioleau: I don't know yet what solution we will be using to get this `#include <data_types/Files.hpp>`.

src/multipatch/connectivity/CMakeLists.txt
tpadioleau: I was treating geometries in the same way because they are in conflict with each other for some header files not because they are conceptually linked.

EmilyBourne: Ok. I had in mind that if we moved to using paths in include statements then conceptually similar files would be treated similarly. However we can also leave this until we actually want to change the includes

src/data_types/CMakeLists.txt
EmilyBourne: Then let's leave this until we decide we definitely want to make this change

tpadioleau:
requires #755 

Should we deprecate old CMake targets instead of simply removing them ?

EmilyBourne:
> Should we deprecate old CMake targets instead of simply removing them ?

I didn't know that was possible! It would probably be cleaner but this is already a giant PR. Adding deprecations would increase the size then require a second giant PR to remove all those deprecations in a version or two

src/advection/CMakeLists.txt
EmilyBourne: @tpadioleau do you think this conversation can be closed? If so I can approve the PR

tpadioleau: I think so, if you agree let's make the installation of Gyselalib++ work so that I can better test Gysela-X++ and the mini app. In a second time we can rework or improve our targets.

EmilyBourne (approved):

tpadioleau:
@EmilyBourne ok to merge ?

PR #767, opened 2026-09-09
v0.8.0

EmilyBourne:


EmilyBourne:
#764 was reverted as the differences between release branch and devel branch were not correctly checked. 3 files should have been deleted

Difference between release branch and devel branch can be confirmed here:
https://github.com/gyselax/gyselalibxx/compare/devel..release

tpadioleau (approved):

